PR #3, opened 2019-03-12
Cvode c merge

EnnaDelfen:
** Updated DVODE integration (still in Fortran)
** New CVODE integration (in cpp)
** + new Fuego cPickler to write improved cpp chemistry routines without rwrk/iwrk. Also, write a few new routines needed for cvode wth fancy integration option
NOTE Tests can be done in the Testing folder. ReactEval is with DVODE, and should sum up the two modifs necessary for retro compatibility. ReactEvalCVODE is same test with CVODE and should show how to use the new chemistry integrator. Tutorials and detailed explanations are on the way.
Note that a regtest does should give 0 errors between ReactEval with DVODE before and after this commit ! It is true with a compare on the dumped pltfiles in ReactEval, and an actual RegTest is on the way.


drummerdoc:
Can you verify that the PeleC tests work with this merge?

AnneFelden:
I can tell you they are not going to work as is, since I've added an argument to actual_react_init. I'm running regtests on pelephysics to make sure everything is OK there, and then I'll generate a branch in PeleC to be consistent with my changes if that sounds ok ?

drummerdoc:
Sure!  Note that there seems to be a need for cvode to be installed in order for the PeleC regtests to work.  That could be a decision we make as a project, or the cvode inclusion could be conditional at compile time, based on a makefile define that defaults to FALSE.

Might be good to get project input for that decision.  If the conditional compiler stuff would be horrendous, that would be useful to know as well.

drummerdoc:
The next two things that are missing are:
1. Documentation of what you've added (to go along with documenting the connection between PelePhysics and PeleC that is still largely empty)
2. Going through and making all the Models compatible with the new stuff (I.e., running Fuego to regenerate all the models so that they would function correctly if selected - not just LiDryer).

(sorry...)

PR #7, opened 2019-03-26
Add 3spmodels to pele physics take2

marient:


marient:
I'll be taking off for vacation next week. Is there anything that I should do to facilitate the merging?

PR #10, opened 2019-04-03
chemistry reordering to increase cache performance

hsitaram:
Some new functions are added in fuego to reorder chemistry so that the memory performance can be improved. Reactions are clustered in a way that the production rate calculation reuses the same/similar species data. The species are reordered in a way that reactions tend to use species ordered continuously in memory. 

We achieve this by constructing a reaction - species matrix and using a traveling salesman reorodering of the rows and columns.

Right now, this reordering has not been turned off in the cpickler (there is a flag that can be used to turn it on) by default and the fuego generated files in Mechanism/models have not been changed.

Also adding a picture to explain this. see pdf below
[rmat_all.pdf](https://github.com/AMReX-Combustion/PelePhysics/files/3040360/rmat_all.pdf)

This shows a matrix representation of the LuDME chemistry (39 spec, 169 reactions). Each row is a reaction and the non-zero (colored) entries are the corresponding species. I have also added all 3rd bodies involved also as a participating specie in a reaction.

Fig 1 from left: the matrix straight out of chemkin
Fig 2 from left: after fuego sorts by reaction type (troe,sri etc..) - this is the DEFAULT
Fig3 from left: after the rows are reordered
Fig4 from left: after columns are reordered

### DIFF
--- a/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py
+++ b/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py
@@ -325,7 +325,7 @@ class CPickler(CMill):
 
     def _renderDocument(self, mechanism, options=None):
 
-        reorder_reactions=True
+        reorder_reactions=False
 
         if(reorder_reactions):
 
--- a/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py
+++ b/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py
@@ -325,10 +325,12 @@ class CPickler(CMill):
 
     def _renderDocument(self, mechanism, options=None):
 
-        plot_react_matrix=True
         reorder_reactions=True
 
         if(reorder_reactions):
+
+            plot_react_matrix = True
+            use_tsp = True #traveling salesman reordering
 
             if(plot_react_matrix):
                 import matplotlib.pyplot as mplt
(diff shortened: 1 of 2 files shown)


drummerdoc:
Any idea how this interacts with the vectorization?  Are there conditions where this is good and conditions where it is a bad thing to do?

hsitaram:
I have seen reduced L2 cache misses when I ran reactEval test case with the grimech; I used intel vtune to measure it. But,  the performance improvement I got was marginal. I am hoping this will have implications on gpu or with larger chemistries. 

So far, with all the cases I have run it has not decreased performance. So its not a bad thing to do. There is also of course a limit on how much one can reorder the chemistry.

AnneFelden:
I might have to take a look at how the routine looks like after this but I believe it won't affect the vectorization. However, it will affect all the pmf's, and it just stresses once again that we should implement an automatic checking of the species order.

hsitaram:
Anne - I have not changed the .cpp files that you had put in. So the test cases should pass.

drummerdoc:
Since this defaults to False, seems reasonable to merge in.  I'm a little worried about how this might impact future work (in that I may need to understand more of this than I do to make sure I don't break it later).  But a good thing to add *soon* is a series of tests to ensure that the reordering doesn't change the answers.  Then if we do something to FUEGO later, we can run those tests to see if the mods need to account for your stuff.  Apparently Jon is doing something that will make this "easy", but a test could be based on the ReactEval in the Testing folder of this repo.

hsitaram:
Thank you Marc. I should also mention that I had done testing with the reordered chemistry and it exactly the same results.

drummerdoc:
You probably mentioned that.  I'm thinking of things that might happen by someone else in the future though.  Having an easy way to test would be a good thing.

hsitaram:
yes. I agree. Although the "diff"ing of the gold regression files will be tricky if the chemistry is reordered.

PR #14, opened 2019-05-03
Fix declaration.

marchdf:
Something was broken when compiling with Null_air


marchdf:
@rgrout and @drummerdoc, can this be merged in?

PR #21, opened 2019-05-31
Add VCKWYR to fuego_chemistry module list.

jrood-nrel:


jrood-nrel:
I will make a corresponding PR for PeleC once this is merged.

PR #22, opened 2019-06-06
Lucas/fix reaction order fuego

esclapez:
Implementing AF changes in Fuego for reaction order of global mechanisms and update of the chem-CH4-2S and JL4.

### DIFF
--- a/Support/Fuego/Mechanism/Models/JL4/JL4.cpp
+++ b/Support/Fuego/Mechanism/Models/JL4/JL4.cpp
@@ -81,7 +81,6 @@
 #define CKEQXP CKEQXP
 #define CKEQYR CKEQYR
 #define CKEQXR CKEQXR
-#define CKCHRG CKCHRG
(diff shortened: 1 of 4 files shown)


EnnaDelfen:
Seems fine. We will have to implement the changes linked to the orders in the Jac too at some point. It's a partial fix for now.

esclapez:
Yes, indeed ! We don't really need the analytical Jacobian of 2-steps schemes right now. But at least we are able to run the new regtests with the master branch of PelePhysics, which seems less confusing (and easier for me to handle in the regression testing *.ini files). We can let the branch open as a reminder ...

drummerdoc:
Made an "Issue" for this topic to try to guard against it getting lost in the noise.

EnnaDelfen:
I'm very proud of us for keeping that... But why wasn't this merged in dev though ?

drummerdoc:
Probably because the person that made this PR chose `master` as the target.

PR #29, opened 2019-06-25
Merge GPU/CPU fuego chemistry files

EnnaDelfen:
All PelePhysics RegTests passed.
I'm just worried about Steven PeleC on GPU. I would like for him to give it a last try and report here. Then I believe it will be safe to merge.
It would be great if everyone doing chemistry-related stuff on GPU could use these cpp routines as a base, so we don't end up scattered all over the place ! No idea what NREL people's requirements are however. If they could have a look too that'd be great.


EnnaDelfen:
PelePhysics regTests all passed. Only LiDryer and DRM19 mechanisms are updated.
Note that the tests performed only checked that the CPU version is not broken.
The GPU version is WIP but it is supposed to be a place to start

drummerdoc:
It all looks good to me, but I don't yet know the impact on NREL or Steven.  You guys gotta weigh in on whether this is save to merge.

EnnaDelfen:
This all started by using a LiDryer mechanism Steven was working with, and he just needed a bunch of them on the GPU. I thought I'd generalize it for all chemistries, but did not necessarily think of generalizing to all routines... I can do that if it's preferable, but maybe it's best to do it incrementally so people can test progressively the new capabilites ? Don't know ...

rgrout:
Ok, either way we can add more later as needed. Other option is that if we don't need them we should suppress outputting them from Fuego to speed up compiling and give compiler less stuff to optimize.

rgrout:
Ok if I squash commits when I merge or would you like to maintain the history separately?

EnnaDelfen:
I believe you can squash away. I have a good idea of what i've changed and we can always go back since I kept everything in Fuego. Thanks

sireeveslbl:
Right now not every function may not be compilable in device code. So not all functions are AMREX_GPU_HOST_DEVICE. We'll need to check the functions so they they do not break any implementations before adding the macro.

PR #32, opened 2019-08-02
Changes to nspecies and nreactions parameter checks to allow for Null Reactions_dir

jrood-nrel:
Maybe I'm doing something wrong, but the PeleC HIT case uses the `Reactions_dir=Null`, which sets these parameters:
```
  integer, parameter :: nspecies    = 1 ! number of species
  integer, parameter :: nreactions  = 0 ! number of reactions
```
which seem valid but they don't currently pass the error checks on the `nspecies` and `nreactions` parameters without these changes.

### DIFF
--- a/Reactions/network.f90
+++ b/Reactions/network.f90
@@ -35,12 +35,12 @@ contains
 
     ! Check to make sure, and if not, throw an error.
 
-    if ( nspecies .le. 1 ) then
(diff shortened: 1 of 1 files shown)


jrood-nrel:
https://travis-ci.org/AMReX-Combustion/PeleC/builds/567052663?utm_source=github_status&utm_medium=notification shows the errors I get for nonreacting cases.

PR #33, opened 2019-08-06
Add missing implicit none statements on subroutines

jrood-nrel:
I noticed some inconsistencies between compilers with PeleC and it looked like it was caused by some missing `implicit none`s so I tried to catch them all in PelePhysics here first.


drummerdoc (approved):
looks good to me

PR #34, opened 2019-08-06
Changing more nspec to nspecies.

jrood-nrel:
I'm pretty lost in everything going on, but I think these need `nspec` to be changed in the development branch as well.


jrood-nrel:
Gah, maybe I can't use `network` in some of these modules for some reason.

drummerdoc (approved):
A lot of those seem unnecessary, and those get_nspecies routines should just be removed, because they turn a parameter into a runtime variable, which seems silly now.  But, it all seems at least correct and consistent.

PR #36, opened 2019-08-28
Feature clean cvode

EnnaDelfen:
Quite self explanatory.
DVODE results should absolutely not change. OMP/MPI safe.
CVODE should now be very easily accessible for every Pele user. To see how, generate the html doc by "make html" in Docs/cvode and follow the instructions to get started.
NOTE: this pull request should not be merged until I make the necessary adjustements in PeleLM and PeleC. The args to react() (cvode and dvode) have (yet again) changed ... Sorry !


EnnaDelfen:
I realize this looks like it is a huge commit. It stems from the fact that the branch I chose to clean things was an experimental one that I progressively cleaned the past few days. Hence the 125 commits, not all of which are actually relevant. I propose that we squash the commits so that it does not pollute the  history of the git repo too much.

PR #40, opened 2019-10-05
Fix preprocessor defines for ionic species

### LINKED ISSUE #39: Naming convention in mechanism.h
Looking at the indices in the generated mechanism.h files, it looks like a few issues might arise. One issue could be when the species have names like "CH2(S)" so the define will have parentheses in it. Another issue arises when the names have a minus (or possibly a plus) like "A4-". In the Aromatic_KrNara directory, the mechanism.h file shows a define statement of A4_ID 117 followed by an A4_ID 118 on the next line. The 117 is supposed to be A4 while the 118 is supposed to be for the A4-. However, it is overwriting the next define.

awehrfritz:
This pull request fixes the preprocessor defines for ionic species, i.e. species names with trailing '-' and '+'.
Also, excited species are now denoted by a lower-case 's' (= star). I can change this to, e.g. 'star', to be more clear or revert it back to 'D', if you think that’s better.

This fixes #39 .

### DIFF
--- a/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py
+++ b/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py
@@ -853,7 +853,15 @@ class CPickler(CMill):
         self._write('/* Species */')
         nb_spec = 0
         for species in mechanism.species():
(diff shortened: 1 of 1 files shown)


drummerdoc:
@EnnaDelfen   What ever happened with this PR?  Can you either accept or reject so we can close this one out? Thanks!

EnnaDelfen:
Cleaned this in dev branch, finally... (see commit 3bcaebe6997d39)

PR #52, opened 2020-01-07
Add safegard on constant transport actual_transport otherwise blocking

esclapez:
when using fpe_trap_overflow.

### DIFF
--- a/Transport/Constant/actual_transport.f90
+++ b/Transport/Constant/actual_transport.f90
@@ -48,10 +48,10 @@ contains
 ! By default the units are in CGS
 ! Below is a flag to let the user to provide MKS data, and it makes the conversion to CGS
     if (mks_unit) then
-      coeff % lam = coeff % lam * 1.0d5
-      coeff % mu = coeff % mu * 10.0d0
-      coeff % xi = coeff % xi * 10.0d0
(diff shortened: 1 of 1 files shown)


drummerdoc:
Not a huge fan of checking for nan's that way, but if it works, I'm good.

esclapez:
Not even checking for nan, just checking that the coeff%mu(:), coeff%xi(:), ... are allocated.

drummerdoc:
oh!  Is that the correct way to check?  Do we need to?  Can we declare it with a def known at compile time and have it always ok?

esclapez:
We can't really do that, it is allocated/deallocated at various time depending on the transport properties you want to compute and the size depends on the size of the box and so on. Just commited a slight change, that is probably more straightforward.

drummerdoc:
Again, didn't know that F90 would understand that code, which looks like you're checking a pointer.

drummerdoc:
Oh...forgot to comment here that I was confused, and that what you did is fine.

drummerdoc (commented):
Looks good to me

EnnaDelfen:
I'm closing this since Emmanuel's commit in dev branch today took care of this !

PR #53, opened 2020-01-29
Anne/clean jac fuego

EnnaDelfen:
This branch has several purposes. 
First, cleaning the name of FUEGO routines so that they be more explicit. 
Second, the implementation of custom sparse solvers using routines dumped by the Gauss-Jordan symbolic solver of DW, both on CPU and GPU
Third the ReactEvalCVODE is cleaned in order to be consistent with what is explained/proposed as a tutorial on the associated readthedoc.
Fourth, cleaning of integration on GPU, here also, to be consistent with readthedoc. NOTE that this last part is WIP, associated readthedoc section should be more descriptive. The test case ReactEvalCVODE_GPU should be operational though. Compatibility with PeleC/PeleLM is not yet assessed.
RegTests are OK.


EnnaDelfen:
Nobody merge this pull request please, CORI is down and I need to verify that I did not forget a commit over there. Also, feel free to express concerns.

PR #54, opened 2020-02-05
Hari/adding chem solvers

hsitaram:
This PR adds ARK ODE chemistry integration with implicit and explicit methods. I have also added feature to use ERK instead of ARK for fully explicit chemistry updates. A test-case called ReactEvalRK has been added to test all RK schemes including the in-house PeleC rk64.


hsitaram:
change LIBRARIES += -Wl,-rpath=${CVODE_LIB_DIR} to LIBRARIES += -Wl,-rpath,${CVODE_LIB_DIR} works on OS x and linux. I reverted because of conflicts in the merge request.

EnnaDelfen:
Hey Hari, thanks for submitting this pull request. 
I have no problem with the modifications of the "actual_CARKODE" routines, since it mainly consists of explicit ODE solvers of use for PeleC. 
That being said, I would like -and I believe it is also what Marc would like- for all these ODE integrators to follow the same "layout". So one thing that bothers me is that you changed the number of arguments to be fed to the "reactor_init" routine for example. As a good test, you might want to make sure 
1) that the actualReactCVODE example could still be properly made to work with any ODE integrators
2) that we can easily switch from calling an integrator in "actual_Creactor" to one in "actual_CARKODE" in any test case you are currently interested with (PeleC, PelePhysics, PeleLM)
I bet this is not possible in the current configuration. So if you really do require additional entries to "reactor_init" or even to "react" routines, then maybe find a way (via optional keywords or compiling instructions variables in the code) to ensure we can compile with all ODE integrators. 

Basically, I believe we should only have one ODE integration test case in Exec. So we should merge ReactEval, ReactEvalCVODE and the one you just created. It is my mistake, I should have done this a while ago (as is, ReactEvalCVODE works perfectly fine with DVODE but I did not remove ReactEval for RegTests purposes).

I believe this will make PelePhysics greater to ensure it is "standalone" and follows a certain structure.

Hope I'm making myself clear ! And if Ray or Marc has any objections, feel free to pipe in.

Anne

EnnaDelfen:
Add on: I also now notice that you coded in the Exec ReactEvalRK an explicit solver. And that makes it confusing. 
Any chance you could make this part of the actual_* family under PelePhysics/Reactions, and make it an option that can be chosen via keywords or at compile time via the GNUmakefile ?
Would be great to have an explicit solver in PelePhysics !

hsitaram:
It should be noted the reactor_init is the same as in the CVODE case except that I have some default arguments. This way the number of required arguments are the same.

hsitaram:
I am in the process writing the CPP version of the explicit case, was hoping to have this pushed in before I send you another large pull request. Likewise, I don't want to lose the fortran version that I had added in the ReactEvalRK folder. That is why I did not delete it.

EnnaDelfen:
I have reviewed the request and attempted to clean a few things. Note that there is now one test case, ReactEval_C where all ODE opt should be avail and ready for testing.
As of now, both CVODE and ARKODE (all 3 options of ARKODE, that is implicit or explicit ARKODE and explicit ERKODE) work on my ignition test case. DVODE too. BUT the Explicit RK solver (select via USE_RK64_PP=TRUE in GNUmakefile) returns nans. 
Could you maybe investigate ? Once we have this we can merge what we have. Regtests for React will be "broken" -- that is, they will not be retrocompatible, but who cares.

Note that I've removed a couple of args to react_init() in arkode and added them in the input_file, to be parsed by the "arkode" parser. That way, all ODE solvers will be similar (I liked the rtol/atol args so I kept them, and will add them to D/CVODE  in the future).

drummerdoc:
Hari, I assume Anne's request was intended for you...I don't have an easy way to debug the failure.  Perhaps you do?

EnnaDelfen:
Sure it was for Hari, it's his explicit RK solver

hsitaram:
I can fix this. It's probably the default tolerances. Thanks a lot of
cleaning up the test cases.

Thanks
Hari

hsitaram:
Ok, I figured out the problem. The rho e was not set correctly. I wonder why all the new "C++" solvers use rho E/rho H for the rhoX argument while just e/h is used for dvode. I think we should change this for being consistent. But that also means we need to change LM and C application codes.

EnnaDelfen:
Hey Hari

If you look into main_nd.F90 in the test case you'll see that if USE_SUNDIALS_PP flag is deactivated then we are supposed to be in old compatible DVODE mode and indeed the energy variable when using the rhoE reactor type is e. I'm thinking now that the way I coded stuff when your explicit solver is enabled, we also should fall into that category, meaning your solver was expected to be fed e and not rhoe... I'm investigating that now

On a general note, in PeleLM we don't use that reactor, we use the reactor type rhoH and dvode/cvode and all solvers should be fed the rhoH variable for energy in this case. All the solvers I have coded are up to date on this. What does it mean ? It means cvode/dvode and arkode can work either with rhoE or rhoH as the energy variable. Now the reason DVODE kept E instead of rhoE is because of the compatibility with PeleC. If you look into Src_nd/React_nd.F90 in PeleC you'll see that rhoE is used only when USE_SUNDIALS_PP flag is activated, exactly like in the ReactEval_C test case. I did not want to mess too much with PeleC !  But if we wanted to completely be consistent, then we would have to modify DVODE to use rhoE and not E for the rhoE reactor type, and modify PeleC to take this into account. It is trivial, I simply did not want to mess with PeleC

hsitaram:
Yeah. I was referring to having dvode get rhoe as the argument instead of e. It could  be a one-line change in pelec but again this will be problem when people use new pelec with old pelephysics (which one ideally shouldn't). Likewise, the problem with RK64 is solved now and hope it will not be problem to merge.

EnnaDelfen:
Yup the RK64 solver works fine now thanks ! 
This merge request seems ready to go...
Brace yourselves people !

drummerdoc:
Oh, before you merge....

PR #55, opened 2020-03-19
C++ EOS, transport, and slightly modified chemistry files

jrood-nrel:
Here are the C++ files for the EOS and transport. Due to MacOS and GNU makefile naming issues, I made `cpp` directories for the new files. For the models, I am requesting we change the filenames to `mechanism.cpp`. I also had to modify the CPickler to add decorators to `CKRHOY` and `CKXTY`. I didn't update every mechanism, but just the ones we've been using so far. I'm sure I missed some things I will need to update later, but this should be everything we needed to get stuff working on the GPU so far.


drummerdoc:
On first pass, it looks good to me. No doubt we'll have to decorate a few more of the functions.  Also, my bet is that there will be a lot of files that can be removed as well once everything is converted over on our side.  

I'd like Anne Felden to look this over as well and see if there are any things she's worried about.

EnnaDelfen:
Thanks for including me in this review. After looking at it a bit more carefully and talking with Marc offline, we do have a couple of comments. Those are more cosmetic issues but in the long run we believe they'll make everything easier and cleaner ! Bear with me because here comes the verbose...

Let's start with the Eos and Transport functions (in Eos/ *** /cpp/EOS.* and Transport/ *** /cpp/*). On a general note I don’t really understand the need for a redeclaration of FUEGO functions inside "extern "C" {}" commands. Everything is already in cpp inside the Fuego (.cpp chemistry files) routines so you should be able to reference those directly. If you are using Fortran, same thing: there is a mod_fuego.f90 inside PelePhysics/Support/Fuego/Mechanism/Models that should enable you to just include a module. 
Specific to the EOS functions, I believe it would be wise to decide right now on a nomenclature. For example, inside the FUEGO chemistry files, you can understand what each function does simply by looking at its name (CK...2...). Right now in the EOS.H/cpp it looks like such a structure has been followed for some selected functions while others have a more cryptic name. As an example, there are two functions "TrhoY2p" and "get_p" which at first glance may seem interchangeable if I don't look at what they do. I don't know if you noticed this when you started cpptizing everything, but there is a file at the root of Eos/Fuego, Fuego_EOS.** where this had already been started as well on our side. Maybe we could fuse our nomenclature ? Also, it would be really nice if the functions that were defined in this file could be included in yours, so as to only have one cpp EOS implementation !Also on a side note -maybe I don't totally understand what you did here: do we really need for the EOS to have a class structure, with an empty init() ?Finally, why the need to redefine already GPUTized FUEGO functions in these EOS files ? As an example, "get_ei" only calls Fuego's CKUMS which is already a host/device function. Can we simplify this and only define new functions that have new purposes ?

If we move into the chemistry modifs, I have no issues with what has been added inside the chemistry routines, or to the CPickler. That looks good. However, can you guys explain the need to have the mechanism.cpp instead of the « real » name (eg mechanism.cpp versus LiDryer.cpp) ? I am not trying to be difficult but if it is just cosmetic, I would advise against it. There are many times when files can be moved switched or modified for whatever reason and I’m afraid it can make things unnecessarily complicated from a bookkeeping perspective. I am under the impression the rename need stems from a Makefile include ? Maybe there is a way around this.If we decide to proceed with this change, I would like to know if that makes a difference in the GNUMakefile of runs ? How do you include a particular chemistry ?  

Carrying this last thought further (sorry) it would be awesome if for both these new EOS and TRANSPORT functions we could have a couple of regtests or at least examples showing the use of these new functions and comparing them to existing F90 ones ! 

Thanks for reading !

jrood-nrel:
Thanks for the response @EnnaDelfen. I will try to number your comments so I can address each of them easier and try not to skip anything:

1. I removed the `extern C` wrappers. Agreed they were unnecessary.

2. We started with an incomplete C++ version of the EOS routines and just built on top of that whenever we needed to fulfill an EOS call. We noticed the `Fuego_EOS.*` files in PelePhysics, but they seemed to be even more incomplete. I agree there should be a standard naming scheme, and we originally tried, but each situation felt too unique for me to be able to create such a scheme on the fly. Now that we have all the calls we need, it might be easier, but I'm going to have to say that I will not be able to provide these names. I would be fine renaming them to anything you prefer if you can provide the list and we can merge the `cpp` files.

3. Personally I would prefer to wrap each direct call to Fuego with an EOS call, because Fuego is quite cryptic in my opinion, and since EOSes are interchangeable in Pele, it seems like there should be some kind of abstraction to make that clear instead of calling Fuego directly. Although, I know PeleC calls certain Fuego routines directly, so I am making a hypocritical statement, but I guess I am just saying that I don't see it being entirely one way or the other.

4. The EOS namespace comes from code which already existed, but I think it works fine as it is. It seems fairly arbitrary to me. It looks like it's trying to resemble the `eos_state` code from the Fortran. I also don't see the empty `init` you are referring to. It looks like `pc_eos_init` got in these files, but I removed it.

5. Using `mechanism.cpp` shows that it is coupled with `mechanism.h` and allows makefiles to be homogenous. Actually I guess it's more coupled with `chemistry_file.H`, so maybe `chemistry_file.cpp` makes more sense. The makefiles are fine with any names for the `cpp` files, and I think the makefiles should probably be lifted out of the directories anyway because they're all just duplicates, but at some point all this feels arbitrary to me as well. Regardless, I only see disadvantages to uniquely naming each mechanism `cpp` file since these models are interchangeable. Actually, I don't really know why there's a `chemistry_file.H` and a `mechanism.h`. Seems like they could be consolidated. In the `GNUmakefile` you choose chemistries with the `Chemistry_Model` argument and it chooses the directory. To me it feels like each model directory should then look almost exactly the same.

6. I agree we should have unit tests for PelePhysics routines, but I do not think they should be compared to the Fortran routines. I think it would be better to have them stand on their own. I enable a fair bit of testing for PeleC and would love to have unit tests for the EOS, transport, etc. However, I won't be able to provide any of that in the scope of this pull request.

With all that said, I don't have any serious convictions to anything I have proposed here. I would usually rather just do whatever necessary to merge things and keep moving. So I think next, we should decide on what should be _required_ to merge this. I think I'm going to agree that we should address (2) before merging, but the rest don't seem as important to me, although fair points. However, that doesn't mean they aren't important if you feel strongly about any of them.

drummerdoc:
Regarding point 5 above, isn't it the case that ``chemistry_file.H`` is duplicated a bunch and should probably be in ``Support/Fuego/Evaluation``?  I put separate ``Make.package`` files in each model folder because ultimately I wanted to properly support that some models may include auxiliary files, such as ``get_rates.cpp`` or ``get_rates.f90``. I thought it was better to make it general enough to specify each model independently.  I'm guessing that one reason you want ``mechanism.cpp`` instead of ``LiDryer.cpp` is that it makes the CMake stuff simpler?  

There are some tests in ``PelePhysics`` already (in ``Testing/Exec``).  Isn't it just a matter of fleshing more of them out?

As for naming scheme, is there significant ambiguity in using: R (rho), T (temp), X (mole), Y (mass), C (molar conc), Q (progress rate), WDOT (production rate), W (molecular wt), Runiv (universal gas const), E (internal energy), H (enthalpy), Cp, Cv, G (gamma), J (Jacobian), JA (approx Jacobian), A (activity), S (entropy).  Then add an I if quantity is a vector with one per species, such as CpI, HI, WI.  Thus, we'd have

eos_T2HI returns a vector of species enthalpies given temperature
eos_RTY2C returns molar concentration given rho, temperature and mass fractions
...

An advantage I see of this is that although the eos_XXX functions are suitably generic to which eos, their names suggest what is needed as input.  If a new eos is added that needs more info, we'll need to remove the old one everywhere and generalize the interface for that function.  But, this way, the compiler can help point out if this happens.  In the unlikely even that we decide pressure should be computed from mass fractions instead of mole fractions, a function called ``get_p`` might pass the compilation and lead to a nasty bug.

EnnaDelfen:
Thanks for both your inputs, glad we agree on most of these points !

I'm answering to your points in order, Jon, including Marc's remark as we go:

1. My bad, maybe I wasn't clear enough: I meant that you could completely remove the declarations of the functions inside that extern C stuff. You should **not** need to declare again the routines that are defined inside the FUEGO mechanisms.cpp, because these are already declared in the chemistry_file.H file ? Am I missing something here ?

2. (and 3.) We all seem to agree this is a crucial point. Here is what I propose we do: 
- we follow Marc's idea for the renaming. It should be a very quick and easy fix. Like this, not only will everyone understand what each function do, but anyone wanting to add functionalities to this file as we go will be able to follow the pattern. We could even add a little comment explaining our naming convention at the top of the EOS.H file. 
- I really think we should remove the wrappers to plain FUEGO functions. I don't agree with them being cryptic in FUEGO, and more than that: FUEGO is at the core of the chemistry and we cannot do without right now. You may have to modify your code a bit but... it will maybe take what, like an hour :) ? And it will make everything so much cleaner ! I count at least 5 superfuous functions that clutter the file for no reason and could totally be removed: get_imw_eos -> get_imw, get_cpi -> CKCPMS, get_wb -> CKMMWY etc.  Of course, someting like "eos_cp" of "get_cp" should be kept even though it may look like a trivial wrapper for some EOS but won't be for more advanced models ! I insist but this really goes against the idea of having a suite of code that is clean and easy to use. Also, the problem of having several functions doing the same thing is obvious: if someone finds a bug, changes the units or whatever in their little piece of code and it brings them to one of these functions I guarantee a fun debug trip for someone further down the line !  
- Once this is dealt with, I will volunteer to put whatever I had coded inside the "Fuego_EOS.**"  and might me missing in your new routines. You are right, it was and is still totally incomplete, but it is used at the moment in a few test cases. So in the interest here also of only having one way of doing similar things ...

4. It's fine by me, I just would like to see an example of how to use these cpp EOS somewhere (regtests ?)
5. Actually Marc the chemistry_file.H is specific to each chemistry because of the arrays' sizes it declares. I never thought of merging the two (chemistry_file.H and mechanism.h) together because mechanism.h is not used most of the time. But I guess we could ! My opinion on this particular matter is less made up but just tell me then: how do you include the chemistry via the GNUmakefile in a specific case, and are you absolutely sure there cannot be a mix up somewhere where the wrong mechanism.cpp is linked ? Thanks.

 6. I agree with Marc: it should only require a few tweaks in already existing test cases. And I don't believe it is something that should be postponed: It is important to make sure that when we compute the Cp, it is the same wether we call "eos_cp" from old fortran or "get_cp" or else from the new cpp.

I believe in moving forward too ! And I believe it can be fairly quick to do so. So I would say if it were **only** up to me, the required things would be to take care of point 2&3, reassure me over point 5, and show me you can at least put point 6 into motion -I can even do this for you, regtests are important.

Thanks !

jrood-nrel:
I will try to address all the requests soon (in my current 10 hours a week work schedule during the pandemic). In the meantime, can you point me to where the PelePhysics test results get reported so I can understand how you're doing the testing?

jrood-nrel:
Ok, I tried to create the requested EOS naming convention. I removed the duplicate declarations for the Fuego routines. I also replaced `get_wb` with `CKMMWY` calls and removed the other unnecessary routines. I originally misunderstood the request to call Fuego routines directly in that you meant from within the Fuego EOS. Where I wanted to minimize calling Fuego routines directly from Pele. I think this covers points 1, 2, and 3.

For point 4, here is how we use the EOS routines in PeleC: https://github.com/AMReX-Combustion/PeleC/blob/7bdfdb917c44b41c38ade8a24af43f6774821856/ExecCpp/RegTests/PMF/prob.H#L179 .

For point 5 the model is chosen in the makefiles like this https://github.com/AMReX-Combustion/PeleC/blob/7bdfdb917c44b41c38ade8a24af43f6774821856/ExecCpp/Make.PeleC#L60 . So I fail to see how the wrong `mechanism.cpp` file would get compiled by accident. I don't see how renaming the files to make them consistent would make this any more likely. We know the headers are all named consistently because we don't want to deal with including differently named files, so why should the `*.cpp` files be any different when we currently have to do the name switch in the makefiles?

For point 6, I haven't seen a response to my question. I do not see any CI connected to this repo, nor can I find where any of the existing tests are running, so I'm hesitant to put any effort into creating tests in which the results don't seem to get reported. If there is a place where these are reported regularly, I will commit to writing the tests in another pull request focused on that. Otherwise I will welcome your offer to write them. I also think unit tests would probably be a better fit for the EOS routines, rather than full regression tests.

drummerdoc:
Hey Jon!  Merry isolation!  Thanks for all the work!

For point 6, long ago we stopped formal CI, and switched to nightly regression tests on a CCSE server with all the other apps.  Then I caused a major breakage in PeleLM and had to pull those tests offline. For no good reason, I pulled PelePhysics ones offline as well.  We are working to fix things and get a formal testing/review process back in place.  Anyway, the testing in PelePhysics has always been of the AMReX type, where local copies of the test results are saved as golden plotfiles, so it's not easily possible to point you to the latest benchmarks - rather you have to generate your own, as usual with the AMReX crap.  The test config file is at PelePhysics/Testing/Regression/PelePhysics-tests.ini

For point 5, how would you handle that some models might have auxiliary implementation files for things like CKWYP that employ QSS approximations, etc?  Not worried about that yet?

jrood-nrel:
Thanks Marc. Why not publish the test results to github like I do with PeleC? https://amrex-combustion.github.io/PeleCRegressionTestResults/ For point 5, I will just revert the file name change.

drummerdoc:
Jon, Yeah, we could do that, but the results are already [posted](https://ccse.lbl.gov/pub/RegressionTesting/) (when actually functioning) with all the other AMReX codes. I also had a scheme to keep the golden results on HSI, but NERSC broke that when they went to MFA.  Then, they were working on a  system connected to the batch system so that you could do regtesting or CI on Cori, but last time I tried, it was still be debugged, and it was frustrating to use (might be better now).

ldowen:
I am confused now. What is the correct way to call chemistry functions? I understand we should call EOS functions like `EOS::TY2Cp` if we want the mixture average specific heat at constant pressure. What if I want the array of specific heats for each species? It looks like there is a function that will get that called `CKCPMS` but this function only seems to exist for Fuego chemistry. So then what should I call if I am using GammaLaw chemistry? Should I add a function to EOS.H that returns the array of specific heats? How should I access the array of molecular weights for each species? I had added these things to the `cpp` branch but they weren't brought over when things switched over in PelePhysics. It would help if the overall methodology of how to call these functions were laid out.

drummerdoc:
The methodology is proposed above, but still not completely fleshed out. I believe that Anne is working on this with Jon so that all the standard cases are covered.  In short, you are correct that you should avoid CHEMKIN/FUEGO-specific calls such as ``CKCPMS`` in favor of the more generic ``EOS::T2CPI`` version, consistent with the comment above.  yet, not all of those exist for you yet.

Maybe you can make a list of the functions that you will need.  You can do this by giving the FUEGO version if you know it, or simply the output and known inputs.  find a reasonably public place to put this list.  Anne should then be able to suggest how to access or create such a function in a way that is applicable to gamma law, Fuego and SRK models.

Sorry that this happens to be all over the floor at the moment, but hopefully it's good that all those invested in this working are engaged in making sure the interface is complete and functional.

ldowen:
I will add my functions to a local branch in PelePhysics and do a pull request to the `cpp` branch. Hopefully, I follow the correct conventions enough that it won't require too much additional work to integrate with the end result. Does this work?

drummerdoc:
At least that will express what you need.  Seems reasonable to me.

EnnaDelfen:
Ok so now I'm confused.
My opinion Marc and Landon, is that we SHOULD refer to CHEMKIN/FUEGO-specific calls EXCEPT when there is a reason to generate a wrapper. I think it is a waste of effort -and can even be dangerous- to over wrap trivial calls such as the molecular weight (get_mw in Fuego). Certain quantities have absolutely nothing to do with the choice of EOS model.

EnnaDelfen:
Ok looking at what Landon commited now I'm over confused. The mw are all equal when using a gamma law ? Does that mean you don't define a Chemistry_dir in your GNUmakefile in your case ? Then how come the variable NUM_SPECIES is defined ? Sorry this probably comes from a lack of understanding of how these other EOS work

drummerdoc:
Although it doesn’t seem like molecular weight has anything to do with thermo or chemistry it does get determined by the “chemistry model” (in the CHEMKIN sense) since that is where the species are identified. I believe also that we should wrap all physics-specific calls (for lack of a better label), including thermo, transport and reactions, in order insulate as much as possible the impact of switching the models. For example it would be extremely useful when using the SRK nonideal model to be able to easily switch to ideal gas or even gamma law in order to see differences and possible to help debug. 

Your worry about unnecessary wrapping is unfounded. The compilers will in-line all that stuff when we finish converting to The GPU style. There is very little reason to call anything like CKCXX directly outside of the PelePhysics implementation files specifically for FUEGO.

jrood-nrel:
Let me know if I need to do anything to help this progress and get merged. Otherwise, I don't have any more commits planned for this and I'm fine with whatever needs to change. I can update PeleC to match anything here after it's settled, but I'm hoping to be able to do that soon.

EnnaDelfen:
Hi all.

Thanks @jrood-nrel for inquiring. I have taken the cpp branch, and put some additional modifications in it that I believe will make everything cleaner for everyone. I have not modified the routines significantly, so your runs should produce the same results. I simply reorganized and renamed some files inside EOS and REACTION folders. I also merged the current dev branch in it.
If you could checkout that branch it is Anne/cpp and try it in your applications that would be great. I believe, if everyone agrees and everyone is happy with this new branch that we should decline this merge request and create another one for that other branch.
NOTE that there is one thing I had to drop is what was inside the EOS.cpp and the last routines in EOS.H. That is because they rely on header files and environment variables only present in PeleC - so a simple test with PP was failing.

Thanks !

ldowen:
So we need to determine where we want to put the `upassMap` and `qpassMap` for PeleC. I can just make it member data in PeleC.H?

EnnaDelfen:
You do need to find another place to define those, yes. pelePhysics is not a good place if it's not shared by all other applications that depend on it !

ldowen:
I have started making the changes to get rid of `upassMap` and `qpassMap`. However, there are function calls like `CKRP` in Eos/GammaLaw/EOS.H but the GammaLaw doesn't have those kinds of functions. Also, would it make more sense to have the universal gas constant, RU, be a pre-processor variable? That way we dont have to call a function that retrieves RU, RU with SI units, and atmospheric pressure when we really just need RU?

jrood-nrel:
Thanks for taking this over @EnnaDelfen . I think all your changes are going in the right direction. So far I can't build until we figure out the `upass` and `qpass` variables in PeleC. For now I don't have any preference where they go @ldowen . So please put them wherever you feel it makes sense. Other developers might have a preference but we can deal with that later. I think I will just close this pull request now and if you can make a new one for `Anne/cpp`, we can continue this work there.

drummerdoc:
@ldowen Our though is to collect up the constants required throughout PelePhysics and put them in a name space that is visible across all models.  We will need to modify FUEGO to refer to these instead of its own set, but that seems straightforward.

PR #56, opened 2020-03-27
Hari/gpu

hsitaram:
This PR adds capability to use SUNDIALS ARKODE integrator for chemistry on GPU. To use, one needs to turn on CUDA (USE_CUDA=TRUE) and turn on ARKODE (USE_ARKODE_PP=TRUE) during build. Current implementation uses a pure explicit source term, there are no linear solves involved. You can also turn explicit RK (ERKStep) by using flag ode.use_erkode=1. 

Preliminary testing on a 16x16x16 grid using drm19 mechanism and the initial conditions from reactval_C test case indicates ~ 40X speed-up (CPU vs GPU). I have verified the results also match for the reacteval_C test case.
cpu_gpu_compare

### DIFF
--- a/Docs/cvode/source/CvodeInPP.rst
+++ b/Docs/cvode/source/CvodeInPP.rst
@@ -570,7 +570,7 @@ To run on GPUs, `AMREX` should be build with CUDA enabled. To do so, add this li
 
     USE_CUDA   = TRUE
 
-Then, the CUDA features of CVODE are enabled in `PelePhysics` via a specific ``USE_CUDA_CVODE`` flag:
+Then, the CUDA features of CVODE are enabled in `PelePhysics` via a specific ``USE_CUDA_SUNDIALS_PP`` flag:
 
 .. code-block:: c++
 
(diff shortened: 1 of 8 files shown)


drummerdoc (commented):
Looks reasonable to me.  The GNUmake include files could be cleaner since the many of the same flags and options are used whether USE_ARKODE_PP is TRUE or FALSE.  Also, I'd eventually like to remove "actual_" from the names of all those files, and make C++ header files .H rather than .h, but what has been done here is consistent with what was there before.

I'd also like to eventually change ``USE_CUDA_CVODE`` to ``USE_CUDA_SUNDIALS`` I think

EnnaDelfen:
Looks good ! I agree with Marc that we should change the CUDA_CVODE stuff into more generic Sundials stuff. I'm going -to be consistent with CPU syntax- to change these to CUDA_SUNDIALS_PP and then adapt the makefile system a little. Then i'll merge. Thanks !

EnnaDelfen:
Ok Hari if it's not too much to ask could you check that these changes don't break your code ? It shouldn't but just to make sure ... I'll merge as soon as you give me the go. Thanks

hsitaram:
I have checked it and its working fine. Thank you for your changes.

PR #63, opened 2020-04-29
some compile error fixes for chemistry integrators

hsitaram:
The new changes to EOS did not percolate into the chemistry integrators. Only the CVODE CPU was updated. I have changed all the CPU and GPU integrators now.


EnnaDelfen:
Hey Hari, you are right I did not take care of the other ODE integrators ! Wanted to but somehow forgot ! Thanks for fixing that this is great 👍  
About the reactEval I saw that you were working on some modifs. To be completely honest I'm not sure I want to pull them in because I would like for the case to be flexible enough to allow the user to directly change these launch options (the ec.stuff) that were here before. We can discuss it however BUT in any case, could you make it so that you propose 2 different branches for merging ? One for the bugs (i e, the modifs of the ODE integrators) and one for that ReactEval ? I know it might ask for more work on your end but we sort of switched to a mode where we want for each specific modifs/improvement to be crystal clear. Thanks again

EnnaDelfen:
There are some good stuff on the ReactEval though, I might pull everything but the actual ParallelFor macro

hsitaram:
Thank you Anne. May be it is better to split it into 2 requests. Let me do
that update you.

hsitaram:
I have changed this PR to having only the fix to the chemistry integrators.

EnnaDelfen:
Awesome thanks ! I'll merge asap :)

PR #64, opened 2020-04-30
Hari/reacteval

hsitaram:
This PR cleans up the reacteval driver. It combines some common operations between CPU and GPU. I have checked it against all the integrators and they all give correct results.


drummerdoc:
Why is this PR closed?

drummerdoc:
Oops, looks like there is something very similar as #65

PR #67, opened 2020-05-09
Feature transport cpp unif

EnnaDelfen:
This only addresses the in/ou in the cpp version of "Simple" and "Constant" transport. It should make it consistent with the F90 versions.
It will affect PeleC


drummerdoc:
Looks good to me

jrood-nrel (approved):
If you want to merge this, I will deal with PeleC changes later. It won't affect us until I update the submodule.

PR #69, opened 2020-05-21
Feature chem cleaning

EnnaDelfen:
This PR cleans the Models folder of PelePhysics in several ways.
1) All chemistry folders now contain the same number of files, and they are all labelled the same: 
        - a tran.dat a therm.dat and a mechanism.inp, which are chemkin input files
        - a make-mechanism.sh to generate Fuego chemistry files
        - a mechanism.h/ chemistry_file.H and mechanism.cpp implementing the Pele chemistry
2) There are a couple of exceptions:
        - The Aromatic_KrNara is still the only gigantic chemistry file and as such has specificities in how the transport is managed. This should be explained in the README contained in its folder
        - The dodecane_wang and drm19 both have extra cpp files, containing routines needed on the GPU to perform a direct solve
3) Not all mechanisms have been updated. There are basically 3 different categories. Updated and tested/ updated but not tested/ deprecated. See the file UpdatedMechanisms.txt for an up-to-date list. I would use the mechanisms in categ 2 and 3 with extreme caution and perform tests to ensure transport and thermo are OK. We might need to implement exhaustive regTests at some point.
4) A couple of handy scripts are in the useful_script folder. In particular the script_generate_mechs.py enables to regenerate all Fuego chemistry files for a given list of mechanisms. The script script_regenerate_make.py will generate a typical chemistryu Make.package


drummerdoc (approved):
This is a rather large PR.  I'm trusting that all has been done well.  Note that we are likely to move all the Fuego generated files to .H to improve inlining and to be consistent with AMReX GPU porting.

drummerdoc:
@awehrfritz and @jrood-nrel I am going to have this PR merged and closed out.  There may be some slight adjustments that the Australian group has to make in order to re-establish their capability with the QSS code in the dodecane folder they use, but it should only be related to new file naming conventions. And NREL can stay on their older branch in their sub repository until ready to move forward.  If I'm wrong on either, please let me know so that I don't screw you up next time as well!

awehrfritz (approved):
@drummerdoc, I am fine with that approach. I have checked the changes in `dodecane_lu` and this looks all good to me.

@EnnaDelfen, for the skeletal mechanism we never used the Fortran file (`dodecane_lu/ckwyp.f`). In this particular case, the Fortran file is actually also just for the skeletal mechanism and does not contain any QSS approximations. 

For the latest runs we switched to a 35-species reduced version (with QSS) for which we then indeed use a Fortran implementation with additional subroutines for QSSA. This subsequently requires rerunning the Fugeo scripts as the species change (35 vs 53 species). Otherwise, we simply add `fEXE_sources+=dodecane_lu_red_getrates.f` to `Make.package` and comment out `CKWC`, `CKWYP` and  `CKWYR` in the C++ file as those functions are then implemented in the Fortran file. This mechanism is not (yet) in the upstream GitHub repo though.

drummerdoc:
Thanks for the input @awehrfritz 

As for the getrates file, that will eventually need to be addressed, as everything gets ported to GPUs, because the strategy is to eliminate all fortran.  It's not as bad as you might think though, because AMReX includes some structs and syntax to make the C++ look a lot like fortran, especially in cases where there isn't a lot of persistent data.  Time frame for that is sorta by the end of the summer or so.

PR #70, opened 2020-05-26
Landon/cpp spray

ldowen:
Fixed some issues with existing functions in EOS.H files and added the `TY2G` function. Changes will need to be made in PeleC to reflect the name change of `YT2Cv` to `TY2Cv`.


Eos/Fuego/EOS.H
```diff
@@ -264,7 +264,8 @@ AMREX_FORCE_INLINE
 void
 EY2G(amrex::Real E, amrex::Real Y[NUM_SPECIES], amrex::Real& G)
 {
-  amrex::Real wbar, Cv, T, Cvx;
+  amrex::Real wbar, Cv, Cvx;
+  amrex::Real T = 300.;
```
drummerdoc: We typically have a good guess for T laying around, and that guess could be passed to this function through another argument.  This seems to be different than the other functions for some reason, no?  Is it worth considering changing this to be EYT2G and passing T?

Eos/Fuego/EOS.H
```diff
@@ -265,9 +253,21 @@ void
 EY2G(amrex::Real E, amrex::Real Y[NUM_SPECIES], amrex::Real& G)
 {
   amrex::Real wbar, Cv, Cvx;
-  amrex::Real T = 300.;
+  amrex::Real T = 500.;
```
drummerdoc, line 256: Here again, we're picking a T and hoping the solver is robust and efficient enough to find the solution, but whenever we call this function, don't we have a decent guess to pass in for T?

EnnaDelfen:
Exactly what Marc said :) + can you actually "rebase" or have a new branch based on dev ? I don't like having those "merge dev into mySuperDuperBranch" ? I don't trust git lately ...

Eos/Fuego/EOS.H
ldowen: I completely agree. This was simply my attempt at keeping the `EY2G` function around. I think getting rid of `EY2G` altogether and make people rely on the `TY2G` function I added would make the most sense. Since gamma needs T and Y, I think we should require those to be passed in.

ldowen:
I can do that. I will cancel this PR and do another one if that is easier for you.

drummerdoc:
Either way would have been fine.  When you add commits to that branch before the PR is merge, they get added automatically to the PR, but a new one is fine too.

PR #71, opened 2020-05-27
Removed EY2G function, added TY2G function, fixed TY2Cv function, add…

ldowen:
…ed missing GammaLaw functions

### DIFF
--- a/Eos/GammaLaw/EOS.H
+++ b/Eos/GammaLaw/EOS.H
@@ -60,7 +60,7 @@ T2Ei(amrex::Real T, amrex::Real Ei[])
 AMREX_GPU_HOST_DEVICE
 AMREX_FORCE_INLINE
 void
-YT2Cv(amrex::Real Y[], amrex::Real T, amrex::Real& Cv)
+TY2Cv(amrex::Real T, amrex::Real Y[], amrex::Real& Cv)
 {
   amrex::Real wbar;
   EOS::Y2WBAR(Y, wbar);
@@ -73,7 +73,7 @@ void
 TY2Cp(amrex::Real T, amrex::Real Y[], amrex::Real& Cp)
 {
   amrex::Real cv;
-  EOS::YT2Cv(Y, T, cv);
+  EOS::TY2Cv(T, Y, cv);
   Cp = gamma * cv;
 }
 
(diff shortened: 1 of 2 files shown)


drummerdoc (approved):
Looks ok to me

ldowen:
I just realized we also need a `RYP2T` function. Should I add that too before completing this PR?

drummerdoc:
Yeah, seems so.

drummerdoc (approved):
looks good

ldowen:
Sorry to keep noting small changes. I am going through PeleC to see what kind of impact these changes will have and noticing other functions that I didn't see before. The `RYET2P` function might have a few issues. Seems like it is actually trying to be `RYE2TP` and energy is passed by reference but should not be modified.

drummerdoc:
Shouldn't R and Y also be passed by reference into RYE2TP and not be modified?

ldowen:
Typically, we pass variables that aren't modified by value and we pass variables that are modified by reference. So the name should be changed to `RYE2TP` and E should be passed by value. Or the function should be eliminated altogether unless it will be needed for the real gas EOS functions.

drummerdoc:
Oh, thanks.  didn't have time to check.  That suggests maybe that @jrood-nrel is using this differently than you think?  Jon, can you comment?

jrood-nrel:
Aren't all variables expected to be pass by reference since it was in Fortran previously?

drummerdoc:
1. I'd prefer everything by reference only because it looks nicer.  But that would probably "break" a lot of code.  We can do that later, maybe?
2. The real question there is how you are using that function `RYET2P` in PeleC.  If Landon makes E read-only will that break your code?

ldowen:
Typically, we would pass everything by reference and just label things that don't change as `const`. However, I assumed there is some issue with doing that in the Fuego routines.

drummerdoc:
There is, as long as we still support F90.

Eos/GammaLaw/EOS.H
```diff
@@ -151,6 +161,15 @@ Y2X(amrex::Real Y[], amrex::Real X[])
...
 
+AMREX_GPU_HOST_DEVICE
+AMREX_FORCE_INLINE
+void
+X2Y(amrex::Real X[], amrex::Real Y[])
+{
+  for (int n = 0; n < NUM_SPECIES; n++)
+    Y[n] = X[n];
+}
+
```
EnnaDelfen, line 182: Hum... Excuse me ? What is that ?

jrood-nrel:
I don't see E getting modified in `RYET2P` so it seems fine to me. I don't know why we are passing it by reference at the moment. Maybe because Fuego expects it? It uses so many pointers. I'm assuming it was an oversight when we moved from Fortran.

drummerdoc:
I think there will some cleanup through all this again once we no longer have a need for F90 code anywhere.  There is still a bunch in PelePhysics and PeleLM for now though.

EnnaDelfen (changes requested):
explain why X=Y please ! Thanks

Eos/GammaLaw/EOS.H
drummerdoc: Gamma law gases don't have molecular weights, so the conversion is nonsensical.  This just ensures that if Sum(X)=1, so does Sum(Y), and preserves all the things that should be preserved.

EnnaDelfen: well there still is a law to convert, X/Xk ends up being 1 but it's still supposed to be valid... But I understand, I did not see this was for the GammaLaw so I guess we're good

EnnaDelfen:
Should we merge ? Everybody good with this ?

drummerdoc:
I'm good.  Merge please

PR #72, opened 2020-06-02
Feature reg tests

EnnaDelfen:
This branch cleans the regTests.
- A CPP version is added for eosEval and TranEval, which should become the only ones once the F90 has been totally removed from all codes.
- The CPP and the FORTRAN are tested in the same folder for ReactEval -which is probably something we should change, since ARKode and the RK solver should also have their own regTests which they do not right now.
- Technically, the CPP versions of all tests should work on GPU. BUT it has not been tested yet since the ODE integration on GPU is currently broken in the development branch (upon which this branch is based). There will be a feature_regTest_GPU branch in the near future to take care of that, once the feature_SUNMatrixCuSolverSP branch has been merged and the ODE integration on GPU fully tested.


Testing/Exec/TranEval/GPU_misc.H
```diff
@@ -45,7 +45,7 @@ initialize_data(int i, int j, int k,
 	rho(i,j,k) = 0.01 + dRho * std::sin(2.0*pi*y/P[1]);
 
 	amrex::Real dummy = 0.0;
-	for (int n = 0; n < NUM_SPECIES; n++) {
+	for (int n = 0; n < NUM_SPECIES-1; n++) {
```
drummerdoc, line 54: What if the last species doesn't have enough mass fraction so that this version ends up putting a negative mass fract there?

drummerdoc (approved):

Testing/Exec/TranEval/GPU_misc.H
EnnaDelfen: It is not possible: if you look at how each mf[n] is constructed they are bounded between Y_hi[n] and Y_lo[n]. The sum of Y_lo[n] (for 1: NUM_SPEC-1) is 0 and of Y_hi[n] is (1- 1.0 / NUM_SPECIES ). So the sum will always be smaller than 1 and we're good.

EnnaDelfen:
OK @drummerdoc I see you have approved this should I go ahead and merge ?

drummerdoc:
sure.  the objection above would only be if you tried to do that in general, but this is a specific test case where we have control over what's happening.  

in general, if you came in with a 3 species problem, and the first two species summed to > 1 for some reason, this would do what I said.  This is the motivation for the "JBB_HACK"...you can say this shouldn't happen, but when subcycling with a base grid that is too coarse, it does happen.

EnnaDelfen:
I am not sure I understand the concern, this is a PP test case, the location of what you commented about is basically an initialization so what you are talking about can never happen. If your concern comes from "is that used in PLM where we have several levels and subcycling and everything else" then the answer is no, there is no need to initialize anything before calling react() in PLM

drummerdoc:
The concern, as you note, is not relevant in an isolated test case, or even in an initialization for PeleLM, but rather that the method might be used later in a different context buried deep inside the reaction chemistry integrator, and that would be a bad thing to do.  The comment here was misplaced.  I typed it too quickly, and it was more a generic statement of which you are already well aware.  All is good.

PR #73, opened 2020-06-03
Feature clean od ekeywords

EnnaDelfen:
instead of specifying the solver to be employed when using CVODE via a number, the user now uses a string.
So dense is cvode.solve_type=dense (and no longer 1), sparse is cvode.solve_type=sparse (and no longer 5) etc. To see the list of options avail, put garbage in the cvode.solve_type and launch your exec.

Included in the PR is also a cleaning of the actual_** in the Reactions/Fuego folder. Note that this will require PR in PLM and PLC as the headers names have changed !


EnnaDelfen:
I actually have a better idea, so wait on the review @drummerdoc

EnnaDelfen:
Ok should be better now. 
All regTests passed.

drummerdoc (approved):
Looks good.  I will generate benchmarks  with this commit, once merged, and will be expecting a companion PR for PeleLM

esclapez (approved):

EnnaDelfen:
yup. The companion is PR 52. Regtests are all green, for PP and PLM.

PR #74, opened 2020-06-08
Add a Y2Wbar Eos function to Fuego.

esclapez:
I assume we don't want to call Fuego routines directly, so this is just a wrapper of CKMMWY.

### DIFF
--- a/Eos/Fuego/EOS.H
+++ b/Eos/Fuego/EOS.H
@@ -141,6 +141,14 @@ TRY2P(amrex::Real R, amrex::Real T, amrex::Real Y[], amrex::Real& P)
 AMREX_GPU_HOST_DEVICE
 AMREX_FORCE_INLINE
 void
+Y2Wbar(amrex::Real Y[], amrex::Real& Wbar)
+{
+  CKMMWY(&Y[0], &Wbar);
+}
+
+AMREX_GPU_HOST_DEVICE
+AMREX_FORCE_INLINE
+void
 RYP2T(amrex::Real R, amrex::Real Y[], amrex::Real P, amrex::Real& T)
 {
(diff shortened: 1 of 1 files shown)


drummerdoc:
Yes, this needs to be done, but don't you need to add one for all the other EOS's?

ldowen:
There is already a function that does that. It is called `Y2WBAR`. We could rename it to be more consistent. It would also have to be changed in GammaLaw. But this PR would mean we have duplicate functions.

drummerdoc:
Ha!  @esclapez gotta do your searches case independent :)

esclapez:
You're right, I did not see this one. I was looking for the chemkin like CKMMWY and did not find it. I can definitely switch to this the one you suggested. Is there a reason why it's not using CKMMWY ?

drummerdoc:
That's a question for @AnneFelden I think

esclapez:
Git blame says @jrood-nrel (not sure tagging here does anything ...)

jrood-nrel:
Do whatever you feel is best.

esclapez:
Okay let's leave it like that and remove the PR/branch. Thanks all!

EnnaDelfen:
If I could put my two cents I'd call the Fuego function. I modified a couple other functions like this one that recomputed what FUEGO is already doing and I thought it best to call that FUEGO function directly because it makes everything consistent. It's sort of a continuation of the cleaning, like when we had 10 defs of Ru and Patm everywhere.

drummerdoc:
I agree with Anne, except to clarify that the Fuego function should not be called if the EOS is not Fuego.  So, the Y2WBAR function should call the CKMMWY function underneath if the EOS if Fuego.

PR #75, opened 2020-06-11
Add TYP2R to wrap CKRHOY: get rho from T, Y and P.

esclapez:
Pretty sure this one did not exist.

### DIFF
--- a/Eos/Fuego/EOS.H
+++ b/Eos/Fuego/EOS.H
@@ -223,6 +223,14 @@ PYT2RE(
 AMREX_GPU_HOST_DEVICE
 AMREX_FORCE_INLINE
 void
+TYP2R(amrex::Real T, amrex::Real Y[], amrex::Real P, amrex::Real &R) 
+{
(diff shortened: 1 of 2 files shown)


Eos/Fuego/EOS.H
```diff
@@ -220,6 +220,14 @@ PYT2RE(
...
 }
 
+AMREX_GPU_HOST_DEVICE
+AMREX_FORCE_INLINE
+void
+TYP2R(amrex::Real T, amrex::Real Y[], amrex::Real P, amrex::Real &R) 
+{
+  CKRHOY(&P, &T, Y, &R);
+}
+
```
drummerdoc, line 230: Don't you need to add this to the other EOS's?

drummerdoc (approved):

Eos/Fuego/EOS.H
esclapez: True, just created a similar function for GammaLaw and actually change the name of the function to be more consistent with the rest of EOS names.

PR #76, opened 2020-06-12
Add GNUmakefile in ThirdParty to manage TP libs

drummerdoc:
**Note that this Pr has a companion PR in [PeleLM #68.](https://github.com/AMReX-Combustion/PeleLM/pull/68) When one is merged, the other needs to be merged as well.**

This feature enhancement adds a GNUmakefile in the ThirdParty folder to manage the build of third-party (TP) libs, SuiteSparse and Sundials, and modifies Make.ThirdParty to set the locations of the libs and includes based on the build config.  The GNUmakefile will control the retrieval of the source code and builds (into a subdirectory "BUILD" in the TP folder) , and install in INSTALL.

To build a CUDA- and KLU-enabled set of sundials and klu libs, for example, you would move to the ThirdParty folder and type ``make USE_CUDA=TRUE USE_KLU=TRUE``.  Then when building an app that included the ``Make.ThirdParty`` file in PelePhysics/ThirdParty that has set those two 'use' variables, this configuration will be selected via the setting of the ``SUNDIALS_LIB_DIR`` variable. Note that unique configs are currently supported based only on DEBUG/CUDA/KLU/COMP. Others can be added later, as necessary.

This can be easily improved to be called (recursively) from the application make system in order to ensure that the config-specific TP libs are made automatically if needed.  Doing this would eliminate the need to document the extra step of building TP libs.  However, there is an issue with the metis build within SuiteSparse, and maybe others, that the new GNUmakefile doesn't seem to properly recognize that some of the dependent libs have already been installed, so if added as a recursive build, it will currently add more work than necessary.  

This is all done as a hack, so the GNUmakefile is probably very brittle, but it seems to work correctly on Godzilla and Garuda.  Further testing and improvements are welcome, including the proper handling of SuiteSparse dependencies to avoid unnecessary re-installs.


ThirdParty/GNUmakefile
```diff
@@ -0,0 +1,95 @@
...
+DEBUG           = FALSE
+USE_CUDA        = FALSE
+USE_KLU         = FALSE
+
+#SS_VERSION=5.7.2
+SS_VERSION=5.4.0
+
+#include $(AMREX_HOME)/Tools/GNUMake/Make.defs
+
+tp_suffix=$(COMP)$(DebugSuffix)$(CUDASuffix)$(KLUSuffix)
```
EnnaDelfen, line 50: Are we sure this is defined here ? most of these $() are not defined yet

ThirdParty/GNUmakefile
```diff
@@ -0,0 +1,95 @@
...
+INSTALL_PREFIX=$(PWD)/INSTALL/$(tp_suffix)
+
+ifeq ($(USE_KLU),TRUE)
```
EnnaDelfen, line 53: I don't understand if suite sparse is build somewhere in there, if does the user have to do it before ?

ThirdParty/GNUmakefile
```diff
@@ -0,0 +1,95 @@
...
+all: $(ONE_SUNDIALS_LIB_FILE)
+
+tp_suffix:
+	@echo $(tp_suffix)
+
+
+$(ONE_SUNDIALS_LIB_FILE): $(ONE_SUNDIALS_SRC_FILE) $(ONE_SS_LIB_FILE)
+	@echo Target $(ONE_SUNDIALS_LIB_FILE)
+	@mkdir -p $(SUNDIALS_BUILD_DIR)
+	cd $(SUNDIALS_BUILD_DIR); cmake -DCMAKE_INSTALL_PREFIX=$(INSTALL_PREFIX) -DCMAKE_VERBOSE_MAKEFILE:BOOL=ON -DCMAKE_C_COMPILER=$(COMP) $(CUDA_HOST_COMP_DEFINE) -DEXAMPLES_INSTALL_PATH=$(INSTALL_PREFIX)/examples -DCMAKE_BUILD_TYPE=$(BUILD_TYPE) -DCMAKE_C_FLAGS_RELEASE=$(COMP_FLAGS) $(CUDA_DEFINES) -DMPI_ENABLE:BOOL=OFF -DOPENMP_ENABLE:BOOL=OFF $(KLU_DEFINES) -DEXAMPLES_ENABLE_C:BOOL=OFF $(SUNDIALS_SOURCE_DIR); make; make install
```
EnnaDelfen: Is there a possibility here to generate a txt file that summarizes the options the user actually chose to build sundials ? I know we will ultimately have the CMakeCache in the builddir but I think it might be good for debug and verification purposes

ThirdParty/GNUmakefile
EnnaDelfen: Ok never mind I god this one

ThirdParty/GNUmakefile
EnnaDelfen: Ok I think I get that one too

ThirdParty/GNUmakefile
EnnaDelfen: I stand by this one though !

EnnaDelfen (approved):

ThirdParty/GNUmakefile
drummerdoc: GNUmake is weird in that it passed through the entire file defining variables before running anything else.

ThirdParty/GNUmakefile
drummerdoc: Accidentally left debugging in.
Re: question. Anything I try to add will be a hacked version of what is actually used.  The CMake record should be sufficient.

emotheau:
I've tested this PR together with the one in PeleLM, it works fine for my environement

PR #77, opened 2020-06-18
fixing some compile issues with arkode for openmp

hsitaram:
arkode integrator was not compiling with openmp. Not sure if the issue is repeated in other integrators.


EnnaDelfen:
Probably just the NLS portion. THe other 3 flags should be the same no matter the thread. Good with me !

EnnaDelfen (approved):

PR #78, opened 2020-06-23
Rename TRY2P to RTY2P

jrood-nrel:
I think this was my fault early on when creating these function names.

Also, can we set the default branch here to `development`? I feel like nobody uses the `master` branch.


EnnaDelfen:
I feel like I've done this already ... Weird ! But yes thanks that looks better :)

PR #81, opened 2020-06-26
Fix ThirdParty builds with the Intel compiler

awehrfritz:
￼


drummerdoc (approved):
Yeah, was doing this on the command line...thanks for fixing it up!

PR #85, opened 2020-07-07
Hari/reactevalfix

hsitaram:
Some fixes to reacteval test case and make system
1) RK64 was not working earlier, rho*e was initialized to just e - fixed
2) CUDA compilation errors fixed - need to use GPUarrays


Testing/Exec/ReactEval_C/GNUmakefile
```diff
@@ -22,7 +22,7 @@ USE_OMP=FALSE
 endif
 
 # define the location of the PELE_PHYSICS top directory
-#PELE_PHYSICS_HOME    := ../../..
+PELE_PHYSICS_HOME    := ../../..
```
drummerdoc: Change this to ``?=``

Testing/Exec/ReactEval_C/GNUmakefile
```diff
@@ -41,12 +46,16 @@ ifeq ($(USE_SUNDIALS_PP), TRUE)
...
+  # Activates GPU versio
+  USE_CUDA_SUNDIALS_PP = TRUE
+  endif
+endif
+
+ifeq ($(USE_RK64_PP), TRUE)
   # Activates use of Hari explicit RK
   # will only work if USE_SUNDIALS_PP = FALSE
-  USE_RK64_PP = FALSE
+  TRANSPORT_CPP = TRUE
```
drummerdoc, line 58: The name of this option makes no sense.

EnnaDelfen: yes it does, it's a flag in PP ( also used in PLM for a while) that enables to choose between transport F90 or CPP ...

drummerdoc: Ok, the +/- here made it look bizarre.

Testing/Exec/ReactEval_C/GPU_misc.H
```diff
@@ -56,11 +56,7 @@ initialize_data(int i, int j, int k, int fuel_id,
 	    frcExt(i,j,k,n) = 0.0;
 	}
 	rhoY(i,j,k,NUM_SPECIES) = temp; 
-#if defined(USE_SUNDIALS_PP)
 	rhoE(i,j,k) = energy * density;
```
drummerdoc, line 59: This one seems very strange.  Is is really the case that the actual evolved variable is (was) different based non whether CUDA SUNDIALS is used?  And if so, is it correct that that is no longer the case??

hsitaram:
All solvers except dvode use rho*e as one of the incoming states

Testing/Exec/ReactEval_C/GNUmakefile
hsitaram: This is done now.

EnnaDelfen:
Note that I am closing this one. The proposed modifs in the def of the GpuArray in main.cpp and misc_GPU.H have already been implemented in development, and PR 87 takes care of the RK solver (should at least).

PR #86, opened 2020-07-08
Feature sunmatrix cusolver sp merge

EnnaDelfen:
This branch fixes the sparse CVODE GPU solver based on the cuSolver (option "cvode.solve_type = sparse"). It had been broken for a while, and I had forgotten about this branch.
The regTests should be OK since the CPU version of reactor.cpp is not affected, other than for a silent modification in the mechanism.cpp files (an option has been added to a sparsity function).
Now we can actually improve the GPU chemistry implementation for all solvers !


drummerdoc:
I don't see anything obviously weird, but this is a lot of code being changed here.

drummerdoc (approved):

EnnaDelfen:
It is a lot of code. Probably will break at some point. Oh well ...

PR #88, opened 2020-07-08
turn off unnecessary packages in sundials build

balos1:
this should speed up the sundials build by disabling unused packages (CVODES, IDA(S), KINSOL)


drummerdoc (approved):
Sure, the less the better!

PR #89, opened 2020-07-09
Update CKSYMS_STR and CKSYME_STR in Fuego.

EnnaDelfen:
That what you guys had in mind ?


ldowen:
Removing the push_back and replacing it with the indexed assignment is what I was thinking. However, I don't think you should change the vectors to arrays or have the function return an array. I think it was fine with the inputs and outputs how they were. I was thinking you would put a `kname.resize(NUM_SPECIES)` at the top of the function CKSYMS_STR function.

ldowen:
I also don't think you can have a function return an array type. Typically, the array should be defined outside of a function and passed in as a pointer.

EnnaDelfen:
@drummerdoc suggested the second point you raise actually. Personnally I'm fine with the way it was or is now but if you or Marc want to make some changes however go ahead :) !

ldowen:
Ok. I will modify it and push it back upstream! Thanks!

drummerdoc:
What's wrong with returning the arrays?

ldowen:
I thought it wasn't allowed in C++.

drummerdoc:
Array's have a (deep) copy constructor, but beside that C++17 std requires that compilers avoid that copy and return the internal object directly (quoted from the master).

drummerdoc:
@ldowen I like the Array's because the sizes are known at compile time and that is made clear (and more efficient) with Array. The push_back's were bad for the reason you say.  And the returning the Array structs is fine, and can be called with ``auto arr = CKSYME_STR();``, so it saves a line of code and a resize call.  The compilers will optimize the implied copy on return to a ref transfer (zero copy).

EnnaDelfen:
Should we merge this ?

ldowen:
My code won't compile with it.

drummerdoc:
Ha! then you're doing it wrong :)
Let me make an example and see what you guys think

drummerdoc:
The simplest form (which wont work on GPU though) is 
`static map<string,Real> awt = { {"a",10}, {"b",20}, {"c",30} };`
Which can be accessed as 
`Print() << "Atomic weight of a is " << awt["a"] << std::endl;`
This avoids all functions...it's just a map.

drummerdoc:
Something that looks closer to what's there is
```
constexpr int Nelts = 3;
static Array<string,Nelts> name = {"a", "b", "c"};
static Array<Real,Nelts> wt = {10, 20, 30};

Array<Real,Nelts>
atomic_weightsCHON()
{
  Array<Real,Nelts> mapped = {0};
  for (int i=0; i<name.size(); ++i) {
    if (name[i] == "c") mapped[0] = wt[i];
    if (name[i] == "b") mapped[1] = wt[i];
    if (name[i] == "a") mapped[2] = wt[i];
  }
  return mapped;
}
```
I'm not a fan of this approach overall.  It's unnecessarily complex and error prone.  But, if this stuff is needed on GPU, the latter version can be written as ``GpuArray`` objects

ldowen:
With the current iteration, I get an error during compilation that the chemistry_file.H has `C-linkage specified, but returns incomplete type`

drummerdoc:
Ahh...why does this need C-linkage? It's not called from Fortran any longer.  Is that needed for device functions?

ldowen:
This is in the reaction mechanism directory, which I think is still called in Fortran (at least, until everything gets ported over).

drummerdoc:
Yes, now I see the issue, and as Anne points out, we're not quite ready to cleave off the Fortran support. So, my suggestions can be delayed.

ldowen:
Ok. Should I modify it as I suggested and push it back? I would just change everything back to vectors and add a resize function call to the start of each relevant function.

drummerdoc:
Yes, I think so.  We can always change out the structures to GpuArrays if needed for GPU application.

EnnaDelfen:
@ldowen I think you should put a totally new PR based on dev to fix this. because ultimately, we are going back and reverting my changes so it's going to be one confusing git history.
Additionally, you only modified GammaLaw but Fuego eos should also have that resize cmd. It would also be good to grep all PP to make sure those **_STR fuego functions are not called elsewhere
Finally, the mechanisms should be changed by adapting the CPickler and not the mechanisms. Then there is a script under useful_script that can regenerate them all afterwards. I can definitely take care of that last part.

ldowen:
So only the GammaLaw EOS has the resize because the resize was added to the `CKSYME_STR` functions, which should handle the Fuego routines. I thought I did do a grep but I could've missed some. Did you find places I missed?

ldowen:
I was also a little confused about something. It seems like the declaration of the function doesn't have a `+sym+` but the definition of the function does. Should this be on both or neither?

EnnaDelfen:
The sym is just an alias for a space so it doesn’t really matter. I know of at least one place it appears and it’s in the reactor cpp in the versions you use. I meant the different eos. Simple le whatever other is in cpp should also be adapted and have a resize of num species. Finally it’s going to impact pele Im too but that’s another issue

ldowen:
In theory, this shouldn't cause any issues anywhere else. Unless other places were set up to use the array format from this PR. The changes should mean the functions can work with both unsized vectors and vectors that are already sized. I have changed it in the development branch and will do a new PR.

PR #91, opened 2020-07-13
Removed push back in reaction mechanisms for vector strings

ldowen:
Update to https://github.com/AMReX-Combustion/PelePhysics/pull/89


drummerdoc (approved):
Looks fine to me, but someone was going to push this to all the other mechs, right @EnnaDelfen ?

PR #93, opened 2020-07-14
Feature typ vals gpu

EnnaDelfen:
Adds typ val and user defined tolerances on the GPU too.


esclapez:
Tested in PelePhysics: check

PR #97, opened 2020-07-16
Clean up omp and printf code, and other misc junk

drummerdoc:
Here, I fixed the arkode reactor in the same way as the code one, and then wen through and cleaned up the omp stuff in the CPU versions to be a little nicer.  There is nothing earthshaking here, but it's a little better.

The only thing functionally different is that you had a place where you only aborted on an error if verbose was on, and that seemed weird.

### DIFF
--- a/Reactions/Fuego/CPU/arkode/reactor.cpp
+++ b/Reactions/Fuego/CPU/arkode/reactor.cpp
@@ -4,6 +4,8 @@
 #include "mechanism.h"
 #include <EOS.H>
 
+using namespace amrex;
+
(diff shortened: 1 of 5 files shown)


EnnaDelfen (approved):
I have no beef with this, but as I'm unsure how it's going to behave with that other branch in which we've made these fixes too (plus others), I'll let you decide if you want to merge Marc.

drummerdoc:
Closing eyes and pushing the button...now fears, Emmanuel will clean up my mess as always... :)

PR #101, opened 2020-07-23
Fixed bug where CPickler did not make the k_r double

ldowen:


drummerdoc (approved):
Odd, how does it even compile without that fix?

EnnaDelfen:
He must be doing something very specific, usually k_r does not appear in tha CPickler generated function. But yes, when we need it we need it declared !

EnnaDelfen:
I am actually curious to know for which situation it lead to an error ? Do you specifi reverses directly somewhere in one of your application ?

EnnaDelfen (approved):

ldowen:
That is correct. I have an application where the mechanism requires specified reverse reaction rates that I tried running on GPUs.

PR #104, opened 2020-07-30
Wrong test for array

drummerdoc:
The Array data, `typVals`, is already allocated, test on the default (nonsense) value of -1 instead.


EnnaDelfen (approved):

EnnaDelfen:
Sure !

PR #106, opened 2020-07-31
Minor mods to arkode CPU

drummerdoc:
Landon, are the these comparable to them mods you made?


ldowen:
Those look like the changes plus some I hadn't seen.

PR #109, opened 2020-08-06
Correct gas constant sig figs

ldowen:
Increased the number of significant digits in R so that it matches what we use in the EOS.


drummerdoc:
There is some nuance here, in terms of which place we want to match significant figures.  @EnnaDelfen has to make this call (and then explain it somewhere)

ldowen:
I did discuss with her, but I would like her to take a look at the PR. What I pointed out was that, if I solve for energy using `PYT2RE` and `RYP2E`, I would get two different values because of the differences in the universal gas constant.

drummerdoc:
Yup, that seems bad. @EnnaDelfen What was the logic used for selecting the current value?

EnnaDelfen:
The current R value, that is, the one that is returned by Fuego's CKRP and used throughout EOS (8.31446261815324e7) comes from Wikipedia. A couple month ago, I ensured we had some level of consistency. What Landon did is go a step further and actually use that same 8.31446261815324e7 value inside the various routines that appear in the mechanism.cpp. So it should be all good !

drummerdoc:
I don't like that the number has to happen to be the same all over.  I think we should define constexpr variable inside the mechanism.cpp that is set from the EOS class data.  Then it only actually appears in one place. Let me poke around for the best way to do that.

EnnaDelfen:
that already kind of is the case, as that R const that appears everywhere in the mechanism.cpp comes from one def inside Fuego CPickler. It was simply not dumped with enough digits. The thing is that this value is also used inside the CPickler to do some internal computations so in my mind having an external constant inside a H file won't solve all problems. In my mind if you really wanted to ensure consistency everywhere you should use that CKRP routine to define your constants, or alternatively dump the CPickler value in the chemistry_file.H and use that in the EOS

drummerdoc:
What's wrong with using ``EOS::RU`` everywhere?  It's known at compile time. We don't actually even need to define anything local to ``mechanism.cpp``

EnnaDelfen:
I thing it's thinking about it backwards. Fuego takes care of chemistry-related constants, and the EOS is the middle man. If you do this, you reverse that logic. Also you can't really use that when launching the CPickler to generate the cpp, you still have to have a definition inside CPickler

drummerdoc:
You can have the single definition location be in CPickler, but then RU should be defined as constexpr in chemistry_file.H.  The definition in EOS.H can remain, but rather set the as a constexpr value using the RU value defined in a namespace appearing in the chemistry_file.H file.  I'm sorry, it's just wrong to have that constant appearing all over the place repeated, whether it's autogenerated or not.

EnnaDelfen:
lol it's not necessarily wrong, especially given what we know about GPU-friendly code. But that's fine if you want to do that !

drummerdoc:
A constexpr variables doesn't "cost" anything, in terms of GPU-friendly code.
The experience you are relying on had to do with "regular" variables and arrays, which were not known at compile time.

drummerdoc:
Hopefully we can remember to "fix" this code correctly using constexpr compile-time expressions

PR #111, opened 2020-08-12
Anne/remove sundials flags

EnnaDelfen:
Remove USE_CUDA_SUNDIALS_PP flag. If USE_CUDA is activated, simply switch to CUDA version of sundials reactor in cvode and arkode.
Also remove TRANSPORT_CPP flag and compile cpp transport by default. IF F90 transport required, then compile those too (the two can coexist)


drummerdoc (changes requested):
In code the define is ``AMREX_USE_CUDA``. Can you make that change throughout?

EnnaDelfen:
oh crap. Sure.

PR #112, opened 2020-08-13
Anne/feature default cpp optional f90

EnnaDelfen:
Only compile CPP files if F90 not required. Makes it so that we can safely remove F90 files in the future !
NOTE: that means that by default DVODE is deprecated


drummerdoc (approved):
the only question is whether it makes sense to have separate F90 flags for all three things...could be ``USE_F90_PP``

EnnaDelfen:
ok sure I can do that too

drummerdoc (approved):

PR #113, opened 2020-08-14
Fix issue with EB:

esclapez:
- integration should be avaoided only for covered cell (mask == -1).
 - For covered cells, set the function count to 0.0 to avoid undefined junk.


EnnaDelfen (approved):

EnnaDelfen:
I wouldn't worry about react_2. Will vanish soon.

drummerdoc (approved):

PR #114, opened 2020-08-31
SRK implementation in C++

baperry2:
Re-implementation of SRK routines from Fortran to C++. Includes some minor algorithmic differences for efficiency. Includes additional functions relative to the Fuego ideal gas EoS where dictated by physics (e.g. REY2T functionally replaces EY2T, which just throws an error for SRK). Also includes some bug fixes in the old Fortran.

### DIFF
--- a/Eos/Soave-Redlich-Kwong/EOS.H
+++ b/Eos/Soave-Redlich-Kwong/EOS.H
@@ -8,20 +8,12 @@
 #include "mechanism.h"
 #include "chemistry_file.H"
 
-/* List of Speedups
- Save oneOverSqrtTc instead of oneOverTc
- Put diagonal of double loops into single loops, only do upper triang of double loop
- Precalculate Amloc and Amlocder
- Parmparse MaxIter and ConvCrit
- Raise error if fail MaxIter or ConvCrit
-*/
-
 namespace EOS {
 
-  // Are these needed for SRK?
   constexpr amrex::Real RU   = 8.31446261815324e7;
-  constexpr amrex::Real RUC  = 1.98721558317399615845;
-  constexpr amrex::Real PATM = 1.01325e+06;
+  // constexpr amrex::Real RU   = 8.3145100000000E+07; // value from old Fortran Version
+  // constexpr amrex::Real RUC  = 1.98721558317399615845;
+  // constexpr amrex::Real PATM = 1.01325e+06;
 
   // Constants for SRK
   constexpr amrex::Real f0 = 0.48508e+0;
@@ -30,12 +22,13 @@ namespace EOS {
   constexpr amrex::Real convCrit = 1e-4;
   constexpr int maxIter  = 2000;
   
-  extern bool initialized;
-  extern amrex::Real Tc[NUM_SPECIES];
-  extern amrex::Real Bi[NUM_SPECIES];
-  extern amrex::Real oneOverTc[NUM_SPECIES];
-  extern amrex::Real sqrtAsti[NUM_SPECIES];
-  extern amrex::Real Fomega[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED extern bool initialized;
+  AMREX_GPU_DEVICE_MANAGED extern amrex::Real Tc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED extern amrex::Real Bi[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED extern amrex::Real oneOverTc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED extern amrex::Real sqrtOneOverTc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED extern amrex::Real sqrtAsti[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED extern amrex::Real Fomega[NUM_SPECIES];
   
 void init();
 
@@ -48,14 +41,12 @@ MixingRuleAmBm(amrex::Real T, amrex::Real Y[], amrex::Real &am, amrex::Real &bm)
 {
   am = 0.0;
   bm = 0.0;
-  amrex::Real Tr ;
+  amrex::Real sqrtT = std::sqrt(T);
   amrex::Real amloc[NUM_SPECIES];
 
   // Combine as follows: add diagonal to am in first loop, loop upper triang in second loop
-  
   for (int ii = 0; ii<NUM_SPECIES; ii++) {
-    Tr = T * oneOverTc[ii];
-    amloc[ii] = (1.0 + Fomega[ii] *(1.0 - std::sqrt(Tr))) * sqrtAsti[ii];
+    amloc[ii] = (1.0 + Fomega[ii] *(1.0 - sqrtT*sqrtOneOverTc[ii])) * sqrtAsti[ii];
     bm += Y[ii] * Bi[ii];
   }
 
@@ -72,14 +63,12 @@ void
 MixingRuleAm(amrex::Real T, amrex::Real Y[], amrex::Real &am)
 {
   am = 0.0;
-  amrex::Real Tr ;
+  amrex::Real sqrtT= std::sqrt(T) ;
   amrex::Real amloc[NUM_SPECIES];
 
   // Combine as follows: add diagonal to am in first loop, loop upper triang in second loop
-  
   for (int ii = 0; ii<NUM_SPECIES; ii++) {
-    Tr = T * oneOverTc[ii];
-    amloc[ii] = (1.0 + Fomega[ii] *(1.0 - std::sqrt(Tr))) * sqrtAsti[ii];
+    amloc[ii] = (1.0 + Fomega[ii] *(1.0 - sqrtT*sqrtOneOverTc[ii])) * sqrtAsti[ii];
   }
 
   for (int ii = 0; ii<NUM_SPECIES; ii++) {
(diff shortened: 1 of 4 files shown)


Eos/Soave-Redlich-Kwong/EOS.H
```diff
@@ -48,14 +41,12 @@ MixingRuleAmBm(amrex::Real T, amrex::Real Y[], amrex::Real &am, amrex::Real &bm)
 {
   am = 0.0;
   bm = 0.0;
-  amrex::Real Tr ;
+  amrex::Real sqrtT = std::sqrt(T);
```
drummerdoc, line 45: Do you want to "protect" this with an assert or something here and below?

Eos/Soave-Redlich-Kwong/EOS.H
```diff
@@ -283,75 +321,73 @@ AMREX_FORCE_INLINE
 void
 TY2Cp(amrex::Real T, amrex::Real Y[], amrex::Real& Cp)
 {
-  amrex::Error("TY2Cp not physically possible for this EoS");
+  amrex::Error("TY2Cp not physically possible for SRK EoS");
```
drummerdoc, line 332: Yeah, I wondered what the best approach to this issue would be.  I'm ok with this strategy for now.  The alternative is that we retrofit all the EOS's with a consistent set of dependencies.

Eos/Soave-Redlich-Kwong/EOS.H
```diff
@@ -363,129 +399,76 @@ EY2T(amrex::Real E, amrex::Real Y[], amrex::Real& T)
...
     // ideal gas internal energy
     CKUMS(&Tn, ei_ig);
     Eig = 0.0; for (int ii = 0; ii< NUM_SPECIES; ii++) {Eig += Y[ii]*ei_ig[ii];};
     // ideal gas heat capacity
     CKCVBS(&Tn, Y, &Cv);
-    //std::cout << "Cv1 " <<Cv << std::endl;
-    //std::cout << "Eig " << Eig << std::endl;
     // real gas heat capacity
-    Cv -= Tn * d2AmdT2 * K1;
+    Cv += Tn * d2AmdT2 * K1;
```
drummerdoc, line 441: I assume this sign change is correct?  I have no idea, but it looked scary enough to call out.

Eos/Soave-Redlich-Kwong/EOS.H
```diff
@@ -496,37 +479,109 @@ RYET2P(
...
-  CKPY(&R, &T, Y, &P);
+  // Note: this function was written to have the same behavior as the Fuego
+  // version (T gets overwritten, so the function is actually RYE2TP)
+  // I'm not sure if this behavior makes sense. The function is only called
+  // once in PeleC, in pc_derpmmserror. One MixingRuleAm could be eliminated
+  // by expanding these function explicitly, but that's small relative to
+  // the overall cost. By using T as in input (unlike Fuego) iteration would
+  // be eliminated and the overall cost would be greatly reduced. - BAP
+  EOS::REY2T(R,E,Y,T);
+  EOS::RTY2P(R,T,Y,P);
```
drummerdoc, line 499: On one hand, the usual comment applies: only waste time on it if it shows up in a profile as significant.  That said, I don't think there's a ton of value in fighting for maximum consistency between the EOS calls.  We can also ifdef code selection based on EOSand make it a compile time decision.

Eos/Soave-Redlich-Kwong/F90/eos_F.F90
```diff
@@ -87,6 +88,14 @@ subroutine actual_eos_init
...
      sqrtAsti(i) = sqrt(Astari(i))
 
+     print *, i-1
+     print *, "Tc", Tc(i)
+     print *, "oneOverTc", oneOverTc(i)
+     print *, "sqrtAsti", sqrtAsti(i)
+     print *, "Fomega", Fomega(i)
+     print *, "Bi", Bi(i)
+     print *, ''
+
```
drummerdoc: Should these be removed, or at least protected behind a verbosity flag?

Eos/Soave-Redlich-Kwong/F90/eos_F.F90
```diff
@@ -418,6 +427,36 @@ subroutine eos_re(state)
...
+ ! print *, "d2AmdT2", state%d2AmdT2
+ ! call Calc_dAmdY(Temp1, state%massFrac, state%dAmdYk)
+ ! print *, "dAmdY", state%dAmdYk
+ ! call Calc_d2AmdTY(Temp1, state%massFrac, state%d2AmdYkdT)
+ ! print *, "d2AmdTY", state%d2AmdYkdT
+ ! call Calc_d2AmdY2(Temp1, state%massFrac, d2AmdY2)
+ ! print *, "d2AmdY2", d2AmdY2
+ ! print *, ""
+
+
```
drummerdoc: These too

Eos/Soave-Redlich-Kwong/F90/eos_F.F90
```diff
@@ -1086,6 +1128,9 @@ subroutine SRK_EOS_Get_T_GivenRhoE(state,lierr)
   ! Update the state structure with the updated Temperature and Pressure
   state%T = Tnp1
 
+  print *, 'T', nIter, Tnp1
+  print *, ''
+
```
drummerdoc: These too.

Eos/Soave-Redlich-Kwong/EOS.H
```diff
@@ -414,13 +414,14 @@ REY2T(amrex::Real R, amrex::Real E, amrex::Real Y[], amrex::Real& T)
   EOS::MixingRuleBm(Y, bm);
   K1 = (1.0 / bm) * log(1.0 + bm*R); 
 
-  // Use input T as initial guess, but constrain it to reasonable bounds
-  Tn = amrex::min(amrex::max(T,300.0),5000.0);
+  // Use input T as initial guess
+  Tn = T;
+  // constrain it to reasonable bounds
+  // Tn = amrex::min(amrex::max(T,300.0),5000.0);
```
drummerdoc: No more protection here?

drummerdoc (commented):
Some questions litter throughout.

Eos/Soave-Redlich-Kwong/EOS.H
baperry2: The next step is to implement this in PeleC so I'll have to decide how to handle this. Option A is to add wrappers like RTY2Cp around TY2CP in other EoS. Option B is leave the interfaces as is and put ifdefs around the EoS calls in PeleC to pick the right interface at compile time. I see advantages and disadvantages to both. I think there is some advantage to having the function interface reflect the physics (Option B), but having a single interface for all EoS is more elegant from a programming perspective. I'd be happy to hear your opinion before starting down one road or the other.

Eos/Soave-Redlich-Kwong/EOS.H
baperry2: The sign change matches other places in the code where Cv is calculated. Here Cv is just used as the derivative in Newton iteration so this bug just slowed down convergence but didn't affect the final result. With the change, iterations to convergence go from ~100 to 5 for some cases.

Eos/Soave-Redlich-Kwong/EOS.H
baperry2: Yeah how this is handled may change eventually. The best strategy may be to just rename the function to RYE2TP so the name better reflects the behavior for each EoS. But right now it seems like a non-issue, so it seems like leaving it in the present form is reasonable.

Eos/Soave-Redlich-Kwong/EOS.H
drummerdoc: I'd say to start the process as you have.  If significant PeleC code needs to be re-orchestrated to make available the data for the enhanced dependency, then if-def'ing becomes impractical (or, at least difficult to read/debug).  The trade-off will then be based on the performance gains of the simpler interface vs. code readability and maintainability.

Eos/Soave-Redlich-Kwong/EOS.H
drummerdoc: If the number of iterations dropped by that much, my guess is that the sign is consistent now with the formulation of the residual that is being driven to zero by the solve, and Jacobian that is being evaluated for it.  5 iterations might well be showing the expected quadratic convergence, whereas 100 can't possibly be in the asymptotic region and so suggests a bug (or a very unfortunate discontinuity at the temperature switch points in the pure Cv fits!).  It would be satisfying to comment out the temperature switches and verify quadratic convergence.

Eos/Soave-Redlich-Kwong/EOS.H
baperry2: Yeah, if you have a negative temperature you're going to have a bad time, checking for that couldn't hurt. I put in some AMREX_ASSERTs to capture this.

Eos/Soave-Redlich-Kwong/F90/eos_F.F90
baperry2: These and other print statements were just a stray remnants of my debugging process. All have been removed.

Eos/Soave-Redlich-Kwong/EOS.H
baperry2: I had second thoughts about hard coding what reasonable bounds are - for example Michael Martin ran a supercritical helium case at ~20 K or something. I think a happy medium here is asserting that the temperature be positive and other than that putting the onus on the user to supply a reasonable guess. I made the corresponding changes in the latest commit.

drummerdoc: The other approach is to ParmParse these in (with sensible defaults) at the beginning of the run.  again, no emergency, but if doing so allows a completely different app to use without hacking code, it seems like a small price to pay.

Eos/Soave-Redlich-Kwong/EOS.H
baperry2: Thanks for the thoughts. In pretty much all cases, the added dependencies for SRK should be available without added effort, so I'm leaning toward the ifdefing option. We'll see if the story changes when I actually sit down to do the implementation though.

drummerdoc (approved):

PR #116, opened 2020-10-15
Cody/extract jac info

balos1:
@EnnaDelfen


Reactions/Fuego/CPU/cvode/reactor.cpp
```diff
@@ -1009,6 +1009,11 @@ int cJac(realtype /* tn */, N_Vector u, N_Vector /* fu */, SUNMatrix J,
           std::cout << std::endl;
       }
       std::cout << "*** END CHEM Jac ***" << std::endl;
+      CVodeGetErrWeights(cvode_mem, tmp1);
+      for (int k = 0; k < N_VGetLength(tmp1); k++){
+        std::cout << N_VGetArrayPointer(tmp1)[k] << " ";
+      }
+      std::cout << std::endl;
   }
```
drummerdoc, line 1017: I don't see a switch to shut this off by default, and code to print in parallel (e.g. Print(), or if IOProc()).
Am I missing something?

PR #117, opened 2020-10-16
Feature null reaction dir

EnnaDelfen:
This branch reinstate the "Null" reaction folder so that when "no chemistry" is needed (in the sense that no reactions occur, not necessarily mean we have no species) then react() simply does nothing.
Makes it possible to compile non reacting test cases without sundials.
Does not break anything or changes any structure in place currently in any of the Pele codes

### DIFF
--- a/Testing/Exec/Make.PelePhysics
+++ b/Testing/Exec/Make.PelePhysics
@@ -71,7 +71,7 @@ ifeq ($(USE_F90_PP),TRUE)
     INCLUDE_LOCATIONS += $(REACTIONS_HOME_F90) $(REACTIONS_PATH_F90)
     VPATH_LOCATIONS   += $(REACTIONS_HOME_F90) $(REACTIONS_PATH_F90)
 endif
-ifdef Chemistry_Model
+#ifdef Chemistry_Model
   ifeq ($(Eos_dir), GammaLaw)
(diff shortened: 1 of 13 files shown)


Testing/Exec/Make.PelePhysics
```diff
@@ -71,7 +71,7 @@ ifeq ($(USE_F90_PP),TRUE)
     INCLUDE_LOCATIONS += $(REACTIONS_HOME_F90) $(REACTIONS_PATH_F90)
     VPATH_LOCATIONS   += $(REACTIONS_HOME_F90) $(REACTIONS_PATH_F90)
 endif
-ifdef Chemistry_Model
+#ifdef Chemistry_Model
```
drummerdoc: Seems weird to have to use defines in GNUmakefiles

drummerdoc (approved):

Testing/Exec/Make.PelePhysics
EnnaDelfen: Yes, I commented them out in case but I'll remove them completely.

EnnaDelfen:
Waiting on Lucas to test it on a PLM case

esclapez (approved):
Looks fine to me. Gives the same results as going through hack_nochem in PeleLM (but HR is not hard zero because of machine precision operations).

PR #125, opened 2020-11-24
Remove Managed Memory from PMF utility.

esclapez:


Utility/pmf.cpp
```diff
@@ -87,25 +85,32 @@ PMF::read_pmf(const std::string& myfile)
...
+    sinput >> pmf_data.pmf_X[i];
+    for (int j = 0; j < pmf_data.pmf_M; j++) {
+      sinput >> pmf_data.pmf_Y[j * pmf_data.pmf_N + i];
     }
   }
 
-  PMF::d_pmf_X = PMF::pmf_X.dataPtr();
-  PMF::d_pmf_Y = PMF::pmf_Y.dataPtr();
+  pmf_data_g = (PmfData *) The_Device_Arena()->alloc(sizeof(pmf_data)); 
+#ifdef AMREX_USE_CUDA
```
drummerdoc: Do we need to add AMREX_USE_HIP conditions here as well?

drummerdoc (approved):

Utility/pmf.cpp
esclapez: That is something I'm not entirely sure. Before we only had NVIDIA/CUDA device, now there are several options, but we still use the USE_CUDA = TRUE and AMREX_USE_CUDA everywhere. The hto2memcpy from AMReX is such that it works regardless, but maybe there is a more generic compiler `ifdef` we should use to see if we compile for GPU ?

drummerdoc: In the Tulip branch that has to be merged, I started the annoying task of doing 
#If defined(AMREX_USE_CUDA) || defined(AMREX_USE_HIP)
or
#If !defined(AMREX_USE_CUDA) && !defined(AMREX_USE_HIP)

We're going to have to propagate those generalize checks here too...

esclapez: Did you have to use a USE_HIP = TRUE in the makefile to have it compile with HIP ?

esclapez: Having a single compiler option to test if CUDA/HIP/DPC++ would be nice, rather than playing with logical combinations of three.

drummerdoc: Yes I did.  And yes, having something like AMREX_USE_GPU would be good. I'll suggest

jrood-nrel: This already exists.

drummerdoc: Hey, at least I guessed correctly :)

PR #127, opened 2020-12-01
Use amrex::Real instead of double during Fuego code generation

jrood-nrel:
I'm not sure if there was a specific reason to use `double`. However, these changes allow me to build PeleC in single precision and may allow for some interesting things in the future.


jrood-nrel:
I should mention I really only did a global replace by word of `double` to `amrex::Real` in `CPickler.py` here and regenerated some mechanisms and everything compiled for me.

PR #124, opened 2020-11-19
bringing arkode integrator to similar code structure as cvode

hsitaram:
This PR adds the array4 integration capability in arkode, both gpu and cpu. It also makes slight modifications to the RK64 to compile correctly. I will add a similar PR for RK64 for completeness in a later PR. Some things to note:

1) I have moved the AMREX_misc.H outside of cvode so that it can be used by all integrators. There exists 2 of them for CPU and GPU, I am sure it can be consolidated. I will defer to Anne/Marc on this.

2) I have unified some of the run time parameters such as rtol and atol under a common rootname for parmparse. It is set to ode now. May be calling it "chemode" will sound better. Again I defer to Anne/Marc on this.

3) The non array4 version for both GPU/CPU may need to be tested in the ReactEval driver as well. I have made sure arkode 1d array integrator works, not done for CVODE.

4) Moving forward I hope we can change the ".h" files to ".H" and change CVODE_BOXINTG flag to just BOXINTG. I have not done that yet, I hope Ann/Marc agree with me.

PR #128, opened 2020-12-12
Fix warnings in CPU arkode reactor.cpp

jrood-nrel:


jrood-nrel:
This is gong to conflict with #124 . Can we merge that?

PR #129, opened 2021-01-07
add interface between AMReX arena and SUNMemory

balos1:
@drummerdoc @AnneFelden @hariswaran


drummerdoc:
Next, need to add the extra initialize line in main.cpp...

Issue #80, opened 2020-06-25
Transport model "Sutherland" is only available through the fortran interface. The C++ is needed for Pele application moving to GPU.


esclapez:

marchdf:
This has been implemented

PR #131, opened 2021-01-12
Feature new fuego

EnnaDelfen:
This PR:
- reorganizes the structure of the chemistry files (chemistry_file.H and mechanism.cpp) to take HIP no managed memory into account. All used and tested mechanisms have been regenerated.
- some unused functions present in the chemistry files have also been removed (but they are still present in the CPickler.py if one ever needs them again)
- All the QSS machinery from the branch julia/QS has been imported in the CPickler but not tested thoroughly yet
- Update the transport to remove managed memory. A pointer to the transport data has to be passed into the routine computing the transport coefficients. Note that issue #80 is resolved with Sutherland transport now available in C++.
- EOS functions are now all AMREX_GPU_DEVICE, i.e. any call to EOS functions from the host will need to be wrapped in a lambda function.
- The Reaction_dir = Null option is now available with GPU.
- A set of CI is implemented to compile/run the Testing/Exec/TranEval case with Constant, Sutherland & Simple transport models on MPI (CPU), MPI+CUDA and HIP.


Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/chemkin/unpickle/mechanisms/QssSpeciesDb.py
```diff
@@ -0,0 +1,44 @@
...
+#
+#                               Michael A.G. Aivazis
+#                        California Institute of Technology
+#                        (C) 1998-2003  All Rights Reserved
+#
+# <LicenseText>
+#
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+
```
drummerdoc, line 13: I wonder if we should not add his licensing stuff...just a thought.

Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/chemkin/unpickle/tokens/__init__.py
```diff
@@ -18,13 +18,14 @@ def sectionsTokenClasses():
 
     from ElementSection import ElementSection
     from SpeciesSection import SpeciesSection
+    from QssSpeciesSection import QssSpeciesSection
```
drummerdoc, line 21: So I understand, are the Qss versions of these files just generalized versions of what was there already so that really, these modified version should replace the original ones rather than add to them?

drummerdoc (approved):

Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/chemkin/unpickle/mechanisms/QssSpeciesDb.py
EnnaDelfen: I hqve no idea. Is it even still valid if it ends in 2003 ?

EnnaDelfen (commented):
@drummerdoc i have no idea what happened to the question about the QSSSpeciesSection but no both this one and the original one needs to be imported. The way it's done, they are both separate objects. 
So if we're all good with this PR I'll go ahead and merge, keeping in mind there is a PLM PR that will follow, and a subsequent PR will deal specifically with GPU and HIP

PR #132, opened 2021-01-13
Fix compile warnings

esclapez:


EnnaDelfen:
Ok so the nukis are supposed to be reals because of the weird 2steps mechanisms were non integer coefs need to be supported. That being said the nuki machinery I believe is not used now that we've reduced the size of the mech.cpp. let me just look into that

PR #138, opened 2021-01-29
Fix memory leak in transport

jrood-nrel:
I don't understand why I couldn't free the host allocations for `trans_parm` within `transport_init` after they have gone through `memcpy` to `trans_parm_g`. I would always get a reference after freed, even though `trans_parm_g` is the only thing referenced after `transport_init`. It seems like it must stay allocated on the host through the duration of the execution. So I chose to make `trans_parm` global. I'm not convinced it's the best solution, but it runs clean like that.

### DIFF
--- a/Transport/Simple/Transport.cpp
+++ b/Transport/Simple/Transport.cpp
@@ -1,11 +1,10 @@
 #include <Transport.H>
 
 TransParm *trans_parm_g;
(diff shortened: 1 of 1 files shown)


jrood-nrel:
I'm going to merge this so I can fix our tests for tonight. Please revert it if necessary.

PR #139, opened 2021-02-08
Hardcode gamma in GammaLaw EOS, at least for now

jrood-nrel:
I say we make this a runtime parameter later, whenever we want to pass `gamma` into all 17 EOS functions that use it. In the meantime, I think it's best to have this stuff working on HIP and DPC++, and this one isn't fun.

### DIFF
--- a/Eos/GammaLaw/EOS.H
+++ b/Eos/GammaLaw/EOS.H
@@ -12,7 +12,7 @@
 
 namespace EOS {
 
-extern AMREX_GPU_DEVICE_MANAGED amrex::Real gamma;
+constexpr amrex::Real gamma = 1.4;
 constexpr amrex::Real RU = 8.31446261815324e7;
 constexpr amrex::Real RUC  = 1.98721558317399615845;
 constexpr amrex::Real PATM = 1.01325e+06;
--- a/Eos/GammaLaw/EOS.cpp
+++ b/Eos/GammaLaw/EOS.cpp
@@ -2,14 +2,9 @@
 
 namespace EOS {
 
-AMREX_GPU_DEVICE_MANAGED amrex::Real gamma = 1.4;
-
 void
 init()
 {
-  amrex::ParmParse pp("eos");
-  pp.query("gamma", gamma);
-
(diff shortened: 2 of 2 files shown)


baperry2:
Ugh, SRK EOS will have the same issue, but we won't be able to hardcode our way out of it.

drummerdoc:
Should we just plan on having a PelePhysics struct that is passed through all of the Eos, transport and chemistry routines?

baperry2:
Perhaps. We possibly need to add additional optional arguments for these routines for reacting LES as well. The additional effort to make both changes at once vs. one or the other alone is likely not too bad.

drummerdoc:
I'll poke around, but it seems like we can bury everything we need for LES, and all the "physics" into a single PelePhysics struct.

esclapez:
I don't think there is a way around it .. but it is kind of a pain. For SRK, is there a way to have everything specified in from the mechanism as for the Fuego EOS ? That would mean that only the GammaLaw EOS need to have a parameter and it doesn't need to be passed into all the reactions pieces (I suppose since gammaLaw + reactions seems weird).
Then we could simply have a couple of overloaded functions to keep the EOS interface the same across the different models.

drummerdoc:
Nah...for constant transport, you have the values. etc.  I think there's no way around it, and modulo some clever idea from Shreyas, I'm thinking we just add a PelePhysics struct to every damn call in PelePhysics (eventually including sprays, LES model parameters, Eos, transport, chemistry, and any other "physics" parameters).

esclapez:
Transport is already setup this way, but it has less ramifications than EOS.

baperry2:
@esclapez Good point on pushing things into the mechanism. For SRK, I do think we could restructure things to have all the parameters we need dumped into the mechanism by Fuego when the mechanism cpp file is generated. It's all stuff that's simple functions of other parameters that already come from the mechanism and will always be the same for any given mechanism.

jrood-nrel:
I'm going to close this because we are working to generalize the EOSes which will account for this.

PR #140, opened 2021-02-17
Nonideal Transport

baperry2:
This pull request contains the corrections necessary to compute transport coefficients with a nonideal EOS (e.g., Chung's model), which have been reimplemented from previous Fortran Code. Other notes about what is included:

- Bug fixes for SRK EOS
- Re-inclusion of the GET_CRITPARAMS function in CPickler.py so it gets included in the mechanism source files when they are generated
- Updating the old Fortran nonideal transport properties code so all constants are double precision (the C++ version then matches to machine precision)
- Since most of the code is the same for ideal and real gas transport properties and the EOS must be selected at compile time anyway, the additional code is turned on/off with an \#ifdef using the flag "PELEPHYSICS_NONIDEAL_EOS"

### DIFF
--- a/Eos/Soave-Redlich-Kwong/EOS.cpp
+++ b/Eos/Soave-Redlich-Kwong/EOS.cpp
@@ -2,14 +2,14 @@
 
 namespace EOS {
 
-  bool initialized=false;
+  AMREX_GPU_DEVICE_MANAGED bool initialized=false;
   
-  amrex::Real Tc[NUM_SPECIES];
-  amrex::Real Bi[NUM_SPECIES];
-  amrex::Real oneOverTc[NUM_SPECIES];
-  amrex::Real sqrtOneOverTc[NUM_SPECIES];
-  amrex::Real sqrtAsti[NUM_SPECIES];
-  amrex::Real Fomega[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real Tc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real Bi[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real oneOverTc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real sqrtOneOverTc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real sqrtAsti[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real Fomega[NUM_SPECIES];
   
 void init() 
 {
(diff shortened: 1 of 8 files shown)


Transport/Simple/F90/actual_transport_nonideal.f90
```diff
@@ -30,8 +30,8 @@ module actual_transport_module
   real(amrex_real), save, allocatable :: Afac(:,:)
   real(amrex_real), save, allocatable :: Bfac(:,:)
   real(amrex_real), save, allocatable :: Sigmaij(:,:)
-  real(amrex_real), parameter :: Avna = 6.022140857e23
-  real(amrex_real), parameter :: PI = 4 * atan (1.0)
+  real(amrex_real), parameter :: Avna = 6.022140857d23
```
drummerdoc, line 33: This is fine here, but should probably be redone across all the code to properly respect double/float compilation flags.

Eos/Soave-Redlich-Kwong/EOS.cpp
```diff
@@ -2,14 +2,14 @@
...
-  amrex::Real oneOverTc[NUM_SPECIES];
-  amrex::Real sqrtOneOverTc[NUM_SPECIES];
-  amrex::Real sqrtAsti[NUM_SPECIES];
-  amrex::Real Fomega[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real Tc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real Bi[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real oneOverTc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real sqrtOneOverTc[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real sqrtAsti[NUM_SPECIES];
+  AMREX_GPU_DEVICE_MANAGED amrex::Real Fomega[NUM_SPECIES];
```
drummerdoc, line 12: @jrood-nrel Is this consistent with DPC++/HIP requirements?

drummerdoc (approved):

Eos/Soave-Redlich-Kwong/EOS.cpp
jrood-nrel: No we can't use these.

jrood-nrel: Best to follow this because I'm going to implement the explicit struct copies in PeleC when I get a chance https://github.com/AMReX-Codes/amrex/pull/1780.

baperry2: Yeah this is just a stop gap (makes it equivalent to what is currently in the code for gamma law), per our conversation over [here](https://github.com/AMReX-Combustion/PelePhysics/pull/139), changes will be needed to make this work for HIP/DPC++.

jrood-nrel (approved):

PR #141, opened 2021-03-03
Templating EOS and Transport

marchdf:
Lots going on in this PR unfortunately but it's hard to keep things contained. At a high level:
- templating and namespacing of EOS
- templating and namespacing of transport
- consolidated GPU and CPU code paths in several mechanisms by changing `CPickler.py`
- consolidated some constants in a given location
- created a `PelePhysics.H` header file which really should be the only thing codes need to include for EOS and transport.

PeleC was updated with these changes and all non-reacting tests pass with zero diffs. Reacting cases lead to machine precision sized diffs so I am pretty confident nothing was broken. Also, this works on non-CUDA GPUs.

The advantage of this PR is that EOS/Transport is now templated on traits so that we can lift a bunch of ifdefs and consolidate all the function calls. For example, in PeleC we had lots of ifdef around SRK calls, now we don't. 

I tried to update PelePhysics itself and its tests. We will see if this PR triggers anything I missed so this is draft for now.

This commit isn't the end of the story... but it's a good stopping place until the next step.


marchdf:
Ok all tests now pass. Ready for review...

jrood-nrel (approved):

drummerdoc (approved):

Transport/Simple/Transport.H
```diff
@@ -11,427 +11,778 @@
...
+                           std::pow(Vcm, 2.0 / 3.0) * B[6] * y * y * H2 *
+                           std::sqrt(Tstar);
+
+    lambda_p *= 4.184e+7; // erg/(cm s K)
+    amrex::Real beta = 1.0 / H2 + B[5] * y;
+
+    // Set nonideal conductivity
+    if (wtr_get_lam) {
+      lam = lam * beta + lambda_p;
+      // lam = H2;
```
baperry2: extraneous commented code left over from testing previous PR, can be deleted

Transport/Simple/Transport.H
```diff
@@ -11,427 +11,778 @@
...
+  AMREX_FORCE_INLINE
+  void operator()(Args...)
+  {
   }
+};
+
+// FIXME remove ifdef when compiling in all the EOS
+#ifdef PELEC_USE_SRK
+template <>
+struct SRKChungCorrections<eos::SRK>
```
baperry2: IMO, this should be called NonidealChungCorrections because the Chung corrections for transport aren't specific to the nonideal EOS

Eos/GammaLaw/EOS.H
```diff
@@ -9,278 +9,324 @@
...
+    RTY2P(R, T, Y, P);
+    dpde = (G - 1.0) * R;
+    dpdr_e = P / R;
+  }
+
+  template <class... Args>
+  AMREX_GPU_HOST_DEVICE GammaLaw(Args...)
+  {
+  }
+  amrex::Real gamma{1.4};
```
baperry2: Would it make sense to specify this in `PhysicsConstants.H` instead?

baperry2 (approved):
One general comment: I think that some combinations of EOS and transport model don't make sense (e.g. GammaLaw and Simple, or SRK and Constant). Is it possible to add error checking to catch these?

Transport/Simple/Transport.H
marchdf: done

Transport/Simple/Transport.H
marchdf: done

Eos/GammaLaw/EOS.H
marchdf: done

sayerhs:
@baperry2 The following code will generate compile-time errors if invalid combinations are attempted. You can add new conditions in future without affecting `PelePhysics` class declaration or any of the code that uses this. In an actual implementation you might want to flip the default to `std::false_type` and explicitly register all allowed types. That will avoid the risk of accidentally forgetting to register disallowed type combinations.

```C++
template <typename EosModel, typename TransportModel>
struct is_valid_physics_combination : public std::true_type
{};

template <>
struct is_valid_physics_combination<GammaLaw, SimpleTransport> : public std::false_type
{};

template <>
struct is_valid_physics_combination<SRK, Constant> : public std::false_type
{};

template <typename EosModel, typename TransportModel>
struct PelePhysics
{
    static_assert(is_eos_type<EosModel>::value,
                  "PelePhysics must have EOS model as its first argument");
    static_assert(is_transport_type<TransportModel>::value,
                  "PelePhysics must have Transport model as its second argument");
    static_assert(is_valid_physics_combination<EosModel, TransportModel>::value,
                  "Invalid physics combination attempted");

    using eos_type = typename EosModel::eos_type;
    using transport_type = typename TransportModel::transport_type;

    static std::string identifier()
    {
        if (std::is_same<EosModel, TransportModel>::value) {
            return EosModel::identifier();
        } else {
            return EosModel::identifier() + "-" + TransportModel::identifier();
        }
    }
};
```

sayerhs:
Sample compiler output 

```
❯ clang++ -std=c++17 test.cpp
test.cpp:81:5: error: static_assert failed due to requirement
      'is_valid_physics_combination<pele::physics::GammaLaw,
      pele::physics::SimpleTransport>::value' "Invalid physics combination
      attempted"
    static_assert(is_valid_physics_combination<EosModel, TransportModel>::value,
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
test.cpp:120:18: note: in instantiation of template class
      'pele::physics::PelePhysics<pele::physics::GammaLaw,
      pele::physics::SimpleTransport>' requested here
    std::cout << PelePhysics<GammaLaw, SimpleTransport>::identifier() <...
                 ^
1 error generated.
```

marchdf:
I will add that before merge!

marchdf:
Adding that stuff is actually going to require building in all the EOSs at the same time. So I am going to hold off until we decide to tackle that. This is a good branching point.

sayerhs:
@marchdf One of the reasons for using type traits is that we can _decouple_ such logic from actual code implementation. You don't need to enable all EOS/Transport types to code... you just need forward declarations telling the compiler that the identifiers used in constraint checks is a type (`struct`). The actual definition is not necessary for the constraint check. You can code up all your constraints in a `PelePhysicsConstraints.H` file with just forward declarations and it never has to see the full definition. You only need the full definition to actually instantiate the `PelePhysics` class. See below for example

```C++
// contents of file test1.cpp
#include <type_traits>
#include <iostream>

namespace pele {
namespace physics {

// Forward declarations
struct Fuego;
struct PengRobinson;
struct GammaLaw;

// Forward declarations
struct Constant;
struct Simple;

// Actual definition 
struct Fuego
{
    using eos_type = Fuego;

};

struct PengRobinson
{
    using eos_type = PengRobinson;

    static std::string identifier() { return "PengRobinson"; }
};

template <typename EosModel, typename TransportModel>
struct is_valid_physics_combination : public std::true_type
{};

template <>
struct is_valid_physics_combination<GammaLaw, Simple> : public std::false_type
{};

} // namespace physics
} // namespace pele

int main()
{
    using namespace pele::physics;

    // Here compiler hasn't seen a definition of GammaLaw or any transport model
    std::cout
        << std::boolalpha
        << "Fuego Constant = "
        << is_valid_physics_combination<Fuego, Constant>::value
        << std::endl
        << "GammaLaw Simple = "
        << is_valid_physics_combination<GammaLaw, Simple>::value
        << std::endl;

}
```

```
❯ clang++ -std=c++17 test1.cpp
❯ ./a.out
Fuego Constant = true
GammaLaw Simple = false
```

marchdf:
cooooool I am on it.

jrood-nrel:
I'm going to merge this. We will need to update PeleLM for this and I will try to help with that as soon as I can.

PR #150, opened 2021-03-16
Updated SootReaction

ldowen:
Just updated the soot reaction.


jrood-nrel (approved):

jrood-nrel:
Actually I have a question. Is it necessary to use `DeviceVector`? It seems like `GpuArray` might suffice since it doesn't need to be a dynamic size.

ldowen:
I just updated the mechanism using the pickler and changed the names of the files. I didn't change anything else. Did I do something wrong?

jrood-nrel:
Oh, I guess I don't know where the `DeviceVector` came from then. @marchdf do you know?

ldowen:
It seems like all of the mechanisms use that.

esclapez:
I think it is coming from a WIP to allow Analytical Jacobian with HIP, so I'm not sure PeleC is using any of it. You can probably open an issue about it to remember that this need further work.

marchdf (approved):

marchdf:
I don't know about this Device Vector stuff... sounds like @esclapez has the right idea though.

PR #153, opened 2021-04-26
Feature fix QSS and Cpickler

EnnaDelfen:
This branch has changed the CPickler.py, fmc.py and LiDryer chemistry files.
The main point of these changes is to accomodate QSS computations on the CPU. The QSS portion is still WIP but it shouldn't impact mechanisms that do not contain QSS !! 
The second point of these changes is to put back all that would be required to do sensitivity analysis inside the CPickler -not dumped in mechanisms file but still there if needed
The third point of these changes is to simplify the mechanisms file by only keeping 2 files: mechanism.H and mechanism.cpp. NO MORE chemistry_file.H !!


esclapez (approved):
I'm good with that PR. Simplifies things on the PelePhysics side and all that is needed in PeleLM is to change the includes of `mechanism.h` into `mechanism.H`.

jrood-nrel (approved):

PR #154, opened 2021-04-29
remove SUNMemory from PelePhysics since it will be in AMReX

balos1:
This requires https://github.com/AMReX-Codes/amrex/pull/1800.


jrood-nrel:
I'm going to close this and deal with it when the amrex PR is merged because this is about to be out of date soon.

balos1:
The AMReX PR has been merged now @jrood-nrel. Should we just reopen it? There does not appear to be conflicts.

PR #156, opened 2021-05-12
Phase 1 to simplifying Reactions directory

jrood-nrel:
This removes `arkode` from the `Fuego/CPU` and `Fuego/GPU` directories. The CPU arkode files are removed and the GPU arkode files are modified to allow for execution on the CPU. This arkode directory is placed directly in the `Reactions` directory. This reduces the file and code complexity at the expensive of a 10% slowdown on the CPU in which I have observed. This single code path will be optimized in the future to hopefully gain this back.

The `Fuego/CPU/RK64` files are moved directly into the `Reactions` directory as well.

Cvode `Fuego/CPU` and `Fuego/GPU` follows similar suit in removing the CPU files and promoting the GPU files to work on both the GPU and CPU at the expensive of capability I believe.  *HOWEVER*, the original `Fuego/CPU/cvode` and `Fuego/GPU/cvode` directories are currently preserved for PeleLM, but I would like to move PeleLM to `Reactions/cvode` in the future.

Ultimately what I'm trying to move towards is a structure similar to the EOSes where we can template reactions in the future in the same manner as the EOSes. So at runtime we should be able to choose from any EOS and any reactions integrator and have minimal code duplication between the CPU and GPU. So:
```
Reactions/arkode/reactions.cpp
Reactions/arkode/reactions.H
Reactions/cvode/reactions.cpp
Reactions/cvode/reactions.H
Reactions/rk64/reactions.cpp
Reactions/rk64/reactions.H
Reactions/null/reactions.cpp
Reactions/null/reactions.H
```
in the end.

`AMReX_SUNMemory*` should go away soon.


esclapez:
Hi Jon ! I'm happy going down this route. My only concern is that the cvode/gpu part might get awfully complex in the near future while testing new options (MAGMA, GINKO, ...).
Also, is it possible to deal with #153 before ?

jrood-nrel:
Ok great. Yeah I'm fine merging #153 first. I can understand cvode becoming more complex, but I think the next step in this involving cvode will make it less complex at first.

esclapez:
@jrood-nrel I can fix the conflicts if you're not on it.

jrood-nrel:
I will fix them tonight.

jrood-nrel:
I say we merge this so I can start using it in PeleC. I'm also anxious to fix whatever makefile mistakes I made with these changes.

esclapez (approved):
Tested with PeleLM without trouble.

jrood-nrel:
Excellent, thanks for checking. Alright I'm merging this.

PR #165, opened 2021-07-15
Move Reactions folder entirely to new organization

jrood-nrel:
This finishes removing the old organization and commits both codes to the new organization in `Reactions`. Note I chose to add a `int reactor_type` argument to `cvode` `react()` calls to avoid using some unacceptable `ifdef` logic in PeleC. Now cvode has one giant `#ifdef AMREX_USE_GPU`, and so does `null`. Arkode is mostly the same code path for GPU and CPU already. `rk64` still isn't used by PeleC yet.

I still have to remove the sun_memory_helper stuff. I may or may not get that done in this PR. I'm going to look at that next. @esclapez let me know if I can answer any questions when you move PeleLM to this.


esclapez:
Thanks ! I'll try to build PeleLM with this branch and see if I run into troubles.

esclapez:
@jrood-nrel I think it's ready to merge. Note that the CPU react() function signature in the null folder has been changed to match that of the cvode one. I'm not sure how that affect PeleC.

jrood-nrel:
Ok great. PeleC doesn't use null (yet at least). I will merge this and we can update each project with the newest commit here at our leisure.

PR #172, opened 2021-07-20
Simplify arkode fkernelspec a bit

marchdf:


hsitaram:
only some aesthetics? we have to move this function outside of arkode/cvode as there is a copy in cvode as well

hsitaram (approved):

jrood-nrel (approved):

marchdf:
yeah just simplifications

marchdf:
@hsitaram I did it for cvode as well. I think we can consolidate later because there's also a CPU version in CVODE.

jrood-nrel:
@hsitaram @esclapez Is this going to mess with anything you're working on? I think we should merge this before it gets out of date.

esclapez:
Yes, I merging now is a good idea.

PR #175, opened 2021-07-21
Enable ReactEval to use actual PeleLM case data.

esclapez:
The user can provide a file containing state and advection/diffusion forcing MFs, as well as a Header with the geometry and boxArray information to perform the chemistry integration.


drummerdoc (commented):
Do you want to indicate maybe that this is assuming a checkpoint file from PeleLM?

esclapez:
It's mentioned at times in the code, but I'll add a comment in the example input file and additional warnings. I was also thinking of Aborting if some tries to use it with energy (instead of enthalpy).
It's not directly a PeleLM checkpoint file, but it assumes that the state is arranged in PeleLM order.

PR #178, opened 2021-07-23
Move reactor h

hsitaram:
removing redundant reactor.H files in cvode and arkode. The routines common to both are now combined


Reactions/arkode/reactor_arkode.cpp
```diff
@@ -1,16 +1,6 @@
-#include "reactor.H"
+#include "../reactor.H"
```
jrood-nrel: Let's not use backwards relative paths for headers.

Reactions/cvode/Make.package
```diff
@@ -1,2 +1,6 @@
-CEXE_headers += reactor.H
-CEXE_sources += reactor.cpp
+#CEXE_headers += reactor.H
+ifneq (, $(filter TRUE, $(USE_CUDA) $(USE_HIP)))
```
jrood-nrel, line 2: I'm reluctant to be on board with this. I really think cvode should be merged into one file like the other integrators. Otherwise a CPU and GPU version of cvode code are going to live forever.

Reactions/reactor.cpp
```diff
@@ -0,0 +1,112 @@
+#include<reactor.H>
```
jrood-nrel: Use `"` for headers we own.

Reactions/reactor.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+using namespace amrex;
```
jrood-nrel: Can we remove this as a default namespace?

Reactions/reactor.cpp
```diff
@@ -0,0 +1,112 @@
...
+
+Array<double, NUM_SPECIES + 1> typVals={-1};
+int eint_rho=1; // in/out = rhoE/rhoY
+int enth_rho=2; // in/out = rhoH/rhoY
+Real relTol=1e-10;
+Real absTol=1e-10;
```
jrood-nrel: Are we fine changing the default arkode tolerance here?

Reactions/reactor.cpp
```diff
@@ -0,0 +1,112 @@
...
+int hack_dump_sparsity_pattern = -5;
+int sparse_custom_solve = 1;
+int use_erkstep = 0;
+int rk_method = 40;
+int rk_controller = 0;
+
+
+//===================================================================
+//Tolerances
+//===================================================================
```
jrood-nrel, line 20: Can we avoid comments like this? Is the `SetTypValsODE` code even involved with tolerances?

ThirdParty/Make.ThirdParty
```diff
@@ -14,10 +14,8 @@ ifeq ($(USE_SUNDIALS_PP),TRUE)
   INCLUDE_LOCATIONS += $(SUNDIALS_LIB_DIR)/../include
   ifeq ($(USE_ARKODE_PP),TRUE)
       DEFINES += -DUSE_ARKODE_PP
-      LIBRARIES += -L$(SUNDIALS_LIB_DIR) -lsundials_arkode
-  else
-      LIBRARIES += -L$(SUNDIALS_LIB_DIR) -lsundials_cvode
   endif
+  LIBRARIES += -L$(SUNDIALS_LIB_DIR) -lsundials_cvode -lsundials_arkode
```
jrood-nrel, line 18: I'm not sure we should be linking to both libraries yet, unless this is a runtime option.

Reactions/reactor.H
```diff
@@ -251,31 +411,25 @@ static int cJac(
...
-void reactor_close();
-
-static int check_flag(void* flagvalue, const char* funcname, int opt);
+//===================================================================
+//Stats and checks
+//===================================================================
+int check_flag(void* flagvalue, const char* funcname, int opt);
 
+//FIXME: make them the same
+#ifdef AMREX_USE_GPU
```
jrood-nrel: Can we use one `#ifdef AMREX_USE_GPU` in this file?

Reactions/reactor.H
```diff
@@ -251,31 +411,25 @@ static int cJac(
...
-  amrex::Real* rhoX_init,
-  amrex::Real* rhoXsrc_ext,
-  amrex::Real* rYs);
-
-// All the following functions are for CUDA only implementation for now
 
+//===================================================================
+// CUDA specific Jacobian functions
+//===================================================================
 #ifdef AMREX_USE_CUDA
```
jrood-nrel: Can we use one `#ifdef AMREX_USE_CUDA` in this file?

Reactions/reactor.cpp
```diff
@@ -0,0 +1,112 @@
...
+        stderr, "\nMEMORY_ERROR: %s() failed - returned NULL pointer\n\n",
+        funcname);
+      amrex::Abort("abort");
+    }
+    return (1);
+  }
+
+  return (0);
+}
+//===================================================================
```
jrood-nrel, line 107: These types of comments seem unnecessary.

jrood-nrel (commented):
In general I support reducing the code by consolidating the similar routines between arkode and cvode, but I'm not sure I'm convinced the code is similar enough yet. It looks like arkode and cvode are attempting to be compiled together. Are you going to make this a runtime option to switch between them? Otherwise, how do the extra cvode routines included in GPU compilation in `reactor.H` succeed for just arkode? Also, can we just move the `reactor_arkode.cpp` and `reactor_cvode*.cpp` files directly into the `Reactions` directory if there is only one file left (in arkode at least) and they will be named differently for each integrator? How do the `null` and `rk64` options fit in here if they do not change?

Reactions/cvode/Make.package
jrood-nrel: This one is fine with me after discussing it.

jrood-nrel (approved):
After discussing this, I'm fine with it as a next step, and we continue to consolidate code within `Reactions`.

Reactions/arkode/reactor_arkode.cpp
hsitaram: I have fixed this now. Was a remnant from when I had the old reactor.H around. Thanks for finding this.

Reactions/reactor.cpp
hsitaram: done

Reactions/reactor.cpp
hsitaram: done

Reactions/reactor.cpp
hsitaram: brought it back to the old values, reltol at 1e-6, in case pelec CI breaks. LM I think is overriding the defaults all the time. @esclapez can you confirm

Reactions/reactor.cpp
hsitaram: typvals are used to set the tolerances, but I removed the comment here and added it at declaration

ThirdParty/Make.ThirdParty
hsitaram: this is necessary as reactor.H uses arkstep and erkstep.h. The next PRs will be to make arkode and cvode selected at runtime. So we will have to compile in both arkstep/erkstep.h and cvode.h

Reactions/reactor.H
hsitaram: I tried to reduce it, sometimes they are needed for arguments. More than one is needed.

Reactions/reactor.H
hsitaram: done

Reactions/reactor.cpp
hsitaram: I added it for readability, which I give priority. I am open to removing if others find this annoying.

Reactions/reactor.cpp
esclapez: Indeed, LM always override with its own values defaulted to 1.0e-10, but it doesn't seem to be a problem for CVODE. So LM is good either way.

PR #179, opened 2021-07-29
Trying to avoid global memory reads

marchdf:


hsitaram:
wish the fkernelspec changes were made once we merge in the previous PR, or @marchdf you could do the same to cvode as well.

marchdf:
This is just for Jon's branch to try out and play with. All that will be
updated before that branch goes to development.

hsitaram:
ahh.. sorry, my bad. should have caught that when CI was not running

marchdf:
Haha

PR #183, opened 2021-08-02
Transpose yvec for global memory coalescence

marchdf:
This ensures that threads have coalesced memory accesses to global memory. It seems to improve `cf_RHS` by 4%. And NCU stops complaining about uncoalesced memory access.


jrood-nrel (approved):

hsitaram (approved):
all good! very clever idea. I wonder what we could do when we want to compile in arkode/cvode together. The dumb way is to have sunvec_index_cvode/sunvec_index_arkode etc..

marchdf:
@hsitaram we do templating! piece of cake ;)

PR #184, opened 2021-08-03
Simplifying the EOS and Transport structures

marchdf:
Curious about what people think of this idea...
- consolidates logic on EOS switches
- compiles in all the EOS (even though we can only pick which one to use at compile time)
- consolidates common EOS functions into one header and cpp file.
- removed all fortran EOS (does anyone still use those?)

The downside is that it requires a bit of rejigging of the make system (eg `Eos_dir` -> `Eos_Model` in the `GNUMakefile`)

If people like this, I might do it for transport too. 

@hariswaran @jrood-nrel @esclapez @drummerdoc, thoughts?


drummerdoc:
Indifferent.  If you can get it to pass all the tests, and if you like it better this way for some reason, I'm good with it

esclapez:
Same here. I don't think anyone is still using Fortran. I'll probably wait until you've updated the transport to make the changes to LM make system.

jrood-nrel (approved):
This is good. This allows us to compile all the EOSes in at the same time and even though switching the EOS at runtime doesn't make much sense, it will allow us to catch more problems, and is one step closer to allowing us to use libraries inside Pele.

marchdf:
Hey everyone, I think this is good to go once the tests pass. LMK what you think.

esclapez (approved):
I'm good with these changes. One step closer to having runtime selection !

Testing/Exec/TranEval_FORTRAN/GNUmakefile
```diff
@@ -19,15 +19,15 @@ PELE_PHYSICS_HOME    := ../../..
 USE_F90_PP = TRUE
 
 ifeq ($(FUEGO_GAS), TRUE)
-  Eos_dir         = Fuego
```
baperry2, line 22: If the Fortran EOS support is being removed, shouldn't the Eval_FORTRAN test cases also be removed?

baperry2 (changes requested):
Looks good. Much easier to follow code organization now. You should also be able to remove the Fortran test cases now.

Testing/Exec/TranEval_FORTRAN/GNUmakefile
marchdf: Good catch @baperry2 . It is done.

baperry2 (approved):

PR #187, opened 2021-08-05
fixes to rk64 integrator

hsitaram:
This PR brings the rk64 integrator to use the same framework as others. It gets compiled in when USE_SUNDIALS_PP is false, USE_RK64_PP is true and Reactions_dir=rk64. It does not need any external libraries, so its useful sometimes. Moving forward we can replace the in-house integrator in pelec to directly use this one.


esclapez (approved):
Looks good Hari ! Do you want to add rk64 testing to the CI ? I can do that and push to this branch if you want.
I was also wondering if we could use the same pointer for both GPU/CPU in ReactEval/main.cpp, but I'm not sure what allocating on The_Device_Arena default to when you are on CPU. I can test that later.

hsitaram:
> Looks good Hari ! Do you want to add rk64 testing to the CI ? I can do that and push to this branch if you want.
> I was also wondering if we could use the same pointer for both GPU/CPU in ReactEval/main.cpp, but I'm not sure what allocating on The_Device_Arena default to when you are on CPU. I can test that later.

thanks Lucas, if you can add the CI for this as well.
About using the same pointer, it was my initial idea until I saw the extra_cells stuff. It would nt be clean if I combine them for CPU/GPU even though extra_cells are going to be 0 for ode_ncells=1

esclapez:
@hariswaran it looks like I'm missing rkparams.H to get the ReactEval case compiling in CI. It's listed in the Make.package, so I think you just forgot to commit it the the repo.

hsitaram:
sorry about that! it should be in now

marchdf:
@baperry2 can you take a look at this and see what would need to be updated for this to work for non-ideal/SRK if we switch to this in PeleC?

Reactions/rk64/reactor_rk64.cpp
```diff
@@ -0,0 +1,296 @@
...
+      for (int i = 0; i < neq; i++) {
+        if (fabs(error_reg[i]) > max_err) {
+          max_err = fabs(error_reg[i]);
+        }
+      }
+
+      if (max_err < captured_abstol) {
+        change_factor =
+          betaerr_rk64 * pow((captured_abstol / max_err), exp1_rk64);
+        dt_rk = std::min(dt_rk_max, dt_rk * change_factor);
```
jrood-nrel: Let's use `amrex::min<amrex::Real>` and `max`.

Reactions/rk64/reactor_rk64.cpp
```diff
@@ -0,0 +1,296 @@
...
+      }
+
+      if (max_err < captured_abstol) {
+        change_factor =
+          betaerr_rk64 * pow((captured_abstol / max_err), exp1_rk64);
+        dt_rk = amrex::min(dt_rk_max, dt_rk * change_factor);
+      } else {
+        change_factor =
+          betaerr_rk64 * pow((captured_abstol / max_err), exp2_rk64);
+        dt_rk = std::max(dt_rk_min, dt_rk * change_factor);
```
jrood-nrel: These `std::max` need to be changed as well. We also need them to be `amrex::max<amrex::Real>` just in case we ever want to use single precision in the future.

Reactions/reactor.H
```diff
@@ -141,20 +152,20 @@ void reactor_close();
...
-  realtype** JSPSmat = NULL;
+  amrex::Real** JSPSmat = NULL;
 #ifdef USE_KLU_PP
   klu_common* Common;
   klu_symbolic** Symbolic;
   klu_numeric** Numeric;
 #else
-  realtype**(**Jbd);
-  realtype**(**P);
+  amrex::Real**(**Jbd);
```
jrood-nrel: Let's also remove the parentheses on these statements since they generate compiler warnings.

Reactions/rk64/reactor_rk64.cpp
hsitaram: done

Reactions/rk64/reactor_rk64.cpp
hsitaram: done

Reactions/reactor.H
hsitaram: @esclapez - these are used in cvode cpu case. I am still baffled why you need a 4d pointer for a 2d matrix?
likewise, I think this should be related to a different PR as this is part of cvode.

esclapez: I have no idea. But that's the form Cvode is expecting ... Probably help minimize the storage of the sparse matrix.

jrood-nrel (approved):

baperry2:
> @baperry2 can you take a look at this and see what would need to be updated for this to work for non-ideal/SRK if we switch to this in PeleC?

I believe the minimum changes to get this to work for SRK are straightforward, and, with the recent simplification of the chemistry solvers, could also be extended for the sundials solvers as well (with the exception of the analytical Jacobian, which @shashankNREL has indicated would be an entire research project in its own right). 

Now that they all share an [fKernelSpec](https://github.com/AMReX-Combustion/PelePhysics/blob/a758e6a8062800983ef12127a9e9eb291bd5485a/Reactions/reactor.H#L325), it should be as simple as modifying the EOS calls there to the SRK compatible interface, e.g. converting eos.EY2T to eos.REY2T. Similar modifications would be needed in `reactor_rk64.cpp` where the EOS is called. For Sundials support, I think the only additional change that would be needed is making the EOS interface changes to the `box_flatten` and `box_unflatten` functions. 

One limitation is that we only have eos.REY2T implemented, so SRK comparability would be limited to constant volume reactors. We'd have to leave the calls like eos.HY2T as they are. But that is all we need for PeleC anyway.

PR #191, opened 2021-08-18
EOS calls for Reactions modified to be SRK-compatible

baperry2:
This pull requests modifies the EOS calls used in the various chemical reaction integrators in PelePhysics to be compatible with real gas equations of state. This should enable PeleC with SRK EOS to use any of the PelePhysics integrators, with the exception of CVODE with analytical jacobian. Also, SRK can only be used with constant volume chemistry integration for now because the RHY2T function has not been implemented for that EOS yet.

So far, integration based on T is still used for SRK.  This involves some additional EOS calls that would not be required for energy-based integration (which may have nontrivial computational cost for SRK). I plan to do some testing to evaluate this before finalizing the pull request.

The changes here are basically just converting calls like eos.EY2T to eos.REY2T, which simply wraps around the former for Fuego and GammaLaw EOS. A few additional wrapper functions were needed to complete this.

### DIFF
--- a/Reactions/null/reactor.cpp
+++ b/Reactions/null/reactor.cpp
@@ -66,9 +66,9 @@ react(
     amrex::Real T_loc = T_in(i, j, k, 0);
     auto eos = pele::physics::PhysicsType::eos();
     if (reactor_type == eint_rho) {
-      eos.EY2T(energy_loc, Y_loc, T_loc);
+      eos.REY2T(rho_loc, energy_loc, Y_loc, T_loc);
     } else {
-      eos.HY2T(energy_loc, Y_loc, T_loc);
+      eos.RHY2T(rho_loc, energy_loc, Y_loc, T_loc);
     }
     T_in(i, j, k, 0) = T_loc;
     FC_in(i, j, k, 0) = 0.0;
@@ -176,9 +176,9 @@ react(
     amrex::Real T_loc = T_in(i, j, k, 0);
     auto eos = pele::physics::PhysicsType::eos();
     if (reactor_type == eint_rho) {
-      eos.EY2T(energy_loc, Y_loc, T_loc);
+      eos.EY2T(rho_loc, energy_loc, Y_loc, T_loc);
     } else {
-      eos.HY2T(energy_loc, Y_loc, T_loc);
+      eos.HY2T(rho_loc, energy_loc, Y_loc, T_loc);
     }
     T_in(i, j, k, 0) = T_loc;
     FC_in(i, j, k, 0) = 0.0;
--- a/Reactions/null/reactor.cpp
+++ b/Reactions/null/reactor.cpp
@@ -176,9 +176,9 @@ react(
     amrex::Real T_loc = T_in(i, j, k, 0);
     auto eos = pele::physics::PhysicsType::eos();
     if (reactor_type == eint_rho) {
-      eos.EY2T(rho_loc, energy_loc, Y_loc, T_loc);
+      eos.REY2T(rho_loc, energy_loc, Y_loc, T_loc);
     } else {
-      eos.HY2T(rho_loc, energy_loc, Y_loc, T_loc);
+      eos.RHY2T(rho_loc, energy_loc, Y_loc, T_loc);
     }
     T_in(i, j, k, 0) = T_loc;
     FC_in(i, j, k, 0) = 0.0;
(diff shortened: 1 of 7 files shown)


marchdf (approved):

Reactions/null/reactor.cpp
```diff
@@ -176,9 +176,9 @@ react(
     amrex::Real T_loc = T_in(i, j, k, 0);
     auto eos = pele::physics::PhysicsType::eos();
     if (reactor_type == eint_rho) {
-      eos.EY2T(energy_loc, Y_loc, T_loc);
+      eos.EY2T(rho_loc, energy_loc, Y_loc, T_loc);
```
marchdf: REY2T?

Reactions/null/reactor.cpp
```diff
@@ -176,9 +176,9 @@ react(
...
     } else {
-      eos.HY2T(energy_loc, Y_loc, T_loc);
+      eos.HY2T(rho_loc, energy_loc, Y_loc, T_loc);
```
marchdf: This too?

baperry2:
I did some playing around based on the ReactEval test case, converted to look at homogeneous autoignition using drm19 at 1800 K, all on CPU. I'm not sure what to make of the results. 

It seems like there is no significant cost difference per RHS evaluation whether the T term in the RHS is populated or not. However, not using T allows for significantly fewer RHS evaluations to be performed when integrating with CVODE (weirdly, no difference for ARKODE, although I'm not 100% sure I did that correctly. Didn't test RK64). This makes sense as including T, especially with a small atol, should be more conservative. However, not using T was actually more robust in addition to being faster - when using T integration failed for larger timesteps, but not when T was omitted from the RHS. I don't really have a good explanation for this.

Screen Shot 2021-08-18 at 9 33 28 PM

Screen Shot 2021-08-18 at 9 33 17 PM

drummerdoc:
I hate this stuff....

marchdf:
uh... so what do we do? Switch to not using T? Feels like that would be a big deal. I have no experience here...

shashankNREL:
For SRK for sure, using e instead of T will be faster and robust. If @baperry2 test shows that using e works well (and is faster) for Fuego (Ideal gas) then we should shift to e for both ideal gas and SRK.

baperry2:
@shashankNREL I'm not prepared to say we should switch based on these results alone, which is just a single test case uncoupled from CFD. Notably, switching would not be compatible with analytical Jacobians (at least in their present form). I do think it's worth investigating a bit more why CVODE doesn't seem to converge as well when T is included.

I think this pull request is ready to go. The effect discussed above seems to be the same for SRK and Fuego, so whatever change (if any) made in response can be handled separately.

hsitaram:
@baperry2 - can you make changes to box_flatten/unflatten as well?

baperry2:
@hsitaram I already made the corresponding changes to box_flatten/unflatten

baperry2:
Regarding the performance considerations above, it appears that if I set `ode.analytical_jacobian = 1` (which enables preconditioning for GMRES but not the analytical Jacobian calculation for this particular compilation configuration), the performance difference is reversed: using T becomes 10x faster than not using T. But not using T and `ode.analytical_jacobian = 0` remains the fastest overall. So this is a situation where the many possible ways of configuring CVODE, including the contained linear solvers,  leads to large and difficult to predict swings in performance, ugh.

marchdf:
Not as part of this PR but... How easy/hard/clean/dirty would it be to enable the capability to switch between the two formulations?

baperry2:
> Not as part of this PR but... How easy/hard/clean/dirty would it be to enable the capability to switch between the two formulations?

Probably not hard/dirty at all to get ~90% of the effect. The changes in behavior are basically obtained by setting the T source term to 0 [here in fKernelSpec](https://github.com/AMReX-Combustion/PelePhysics/blob/706c9fbf9a7b26456bae8557128c1cfdb3de2465/Reactions/reactor.H#L358). The only thing that would be a bit annoying would be passing an added argument to fKernelSpec to tell it which method should be used.

hsitaram:
> @hsitaram I already made the corresponding changes to box_flatten/unflatten

sorry I missed that. I think its good to go.

hsitaram (approved):

marchdf:
@baperry2 sounds easy enough. Maybe worth doing in another PR at some point in the future.

PR #192, opened 2021-08-23
Capture udata pointer data explicitly and avoid global memory access

marchdf:
Either the compiler or AMRex decides to copy the object pointed by
udata into GPU global memory and caused the stalls. Our Rice
University collaborators found that this leads to a 7% improvement.

### DIFF
--- a/Reactions/arkode/reactor_arkode.cpp
+++ b/Reactions/arkode/reactor_arkode.cpp
@@ -440,7 +440,7 @@ int cF_RHS(realtype t, N_Vector y_in, N_Vector ydot_in, void* user_data)
 //          udata->rhoe_init_d, udata->rhoesrc_ext_d, udata->rYsrc_d);
 //      }
 //    });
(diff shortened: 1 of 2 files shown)


marchdf:
@jrood-nrel and @esclapez, let's give this a shot on GPU and see if we see a performance gain.

jrood-nrel:
This gives a 4% speedup on our PeleC PMF drm19 test case on the Eagle V100s.

jrood-nrel (approved):

marchdf:
I guess we would want to check that it's still correct... haha

drummerdoc:
If it compiles, it must be correct, right! :)

jrood-nrel:
I didn't do fcompare but stdout is almost exactly the same.

jrood-nrel:
I didn't test cvode if @esclapez can do that.

jrood-nrel:
I tested this with cvode. Gives a slight speedup in PeleC.

PR #195, opened 2021-08-26
Nonideal transport efficiency improvements

baperry2:
Speed up Simple transport for SRK EOS by precomputing:
- square roots in an Nspec^2 loop
- a bunch of multiplications in an Nspec^3 loop (most of the performance impact)

Requires storing an Nspec^3 array in trans_parm, but reduces the cost of transport coefficient calculations by about 2x. For PeleC's PMF-SRK with drm19 on CPU and rk64 chemistry, this leads to ~20% overall reduction in cost. On GPU, the speedup is smaller (a few %), but could be larger if Sundials reduces the cost of chemistry integration. 

Draft until tested more thoroughly

### DIFF
--- a/Transport/Simple.H
+++ b/Transport/Simple.H
@@ -44,29 +44,28 @@ struct NonIdealChungCorrections<eos::SRK>
 
     for (int i = 0; i < NUM_SPECIES; ++i) {
       for (int j = 0; j < NUM_SPECIES; ++j) {
-        const amrex::Real T2 =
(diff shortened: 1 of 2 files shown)


marchdf:
Cool! Can you rebase on `development` so we can see the changes?

marchdf (approved):

baperry2:
This is ready to go now. Fixed a bug and now gives no diff relative to old code.

PR #196, opened 2021-08-27
Add error check for pmf file, add turbinflow files and test code

drummerdoc:
The test code reads a turbulence file and uses it to initialize a MultiFab and write it as a plot file.  If run without specifying the turb_dir, it will create a "turbulence" file first (actually, just a simple 3-component field), write it in the correct format, then read that file and create the output.  Thus, the test demonstrates how to create the data file expected by the turbinflow code, and then how to set up the interpolation stuff to read it and initialize data with it.  This commit adds turbinflow.{H,cpp} to the PeleC build (always).  I also added an unrelated error check to ensure the pmf datafile is properly opened.

### DIFF
--- a/Utility/turbinflow.H
+++ b/Utility/turbinflow.H
@@ -1,0 +1,64 @@
+#ifndef _turbinflow_H_
+#define _turbinflow_H_
+
+#include <AMReX_FArrayBox.H>
+#include <AMReX_GpuMemory.H>
+#include <AMReX_Geometry.H>
+
+struct TurbParmHost
+{
+  amrex::Gpu::DeviceVector<long> offset_dv;
+  std::string turb_file = "";
+};
+
+struct TurbParm
+{
+  amrex::GpuArray<int, AMREX_SPACEDIM> npboxcells = {{0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> pboxlo = {{0.0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx = {{0.0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dxinv = {{0.0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> pboxsize = {{0.0}};
+
+  int nplane = 32;
+  amrex::FArrayBox* sdata = nullptr;
+  amrex::Real szlo = 0.0;
+  amrex::Real szhi = 0.0;
+  bool isswirltype = false;
+  amrex::Real turb_scale_loc = 1.;
+  amrex::Real turb_scale_vel = 1.;
+  amrex::Real turb_conv_vel = 1.;
+  bool turbinflow_initialized = false;
+  bool turbinflow_planes_initialized = false;
+  int kmax;
+  long* offset;
+  long offset_size;
+  TurbParmHost* tph = nullptr;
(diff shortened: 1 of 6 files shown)


Utility/turbinflow.H
```diff
@@ -0,0 +1,64 @@
...
+struct TurbParm
+{
+  amrex::GpuArray<int, AMREX_SPACEDIM> npboxcells = {{0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> pboxlo = {{0.0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx = {{0.0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dxinv = {{0.0}};
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> pboxsize = {{0.0}};
+
+  int nplane = 32;
+  amrex::FArrayBox* sdata = nullptr;
```
marchdf, line 23: Can we `unique_ptr` this so we don't have to worry about issues of deleting, etc?

Utility/turbinflow.H
```diff
@@ -0,0 +1,64 @@
...
+  bool isswirltype = false;
+  amrex::Real turb_scale_loc = 1.;
+  amrex::Real turb_scale_vel = 1.;
+  amrex::Real turb_conv_vel = 1.;
+  bool turbinflow_initialized = false;
+  bool turbinflow_planes_initialized = false;
+  int kmax;
+  long* offset;
+  long offset_size;
+  TurbParmHost* tph = nullptr;
```
marchdf, line 35: same here

Utility/turbinflow.H
```diff
@@ -0,0 +1,64 @@
...
+  amrex::Real szlo = 0.0;
+  amrex::Real szhi = 0.0;
+  bool isswirltype = false;
+  amrex::Real turb_scale_loc = 1.;
+  amrex::Real turb_scale_vel = 1.;
+  amrex::Real turb_conv_vel = 1.;
+  bool turbinflow_initialized = false;
+  bool turbinflow_planes_initialized = false;
+  int kmax;
+  long* offset;
```
marchdf, line 33: unique_ptr?

marchdf (approved):
Looks good. I am just suggesting we remove raw pointers and replace with `unique_ptr` since that is more modern and recommended. Otherwise we have to worry about memory leaks.

Utility/turbinflow.H
drummerdoc: Good point

Utility/turbinflow.H
drummerdoc: This one too.

Utility/turbinflow.H
drummerdoc: Not this one...never allocated. I wonder if there is a safer/cleaner what to point to the offsets.

marchdf: oh I see. hmmmm good question...

drummerdoc: Oh wait, offset is indeed an array of longs, where the long is a seek position.  So offset can be wrapped as well.

esclapez:
@drummerdoc Can we drop this PR ? I think it is obsolete now.

esclapez:
Closing this PR now, it is obsolete.

PR #207, opened 2021-10-08
Update SUNMemory files

jrood-nrel:
Unfortunately it's proving difficult to use AMReX's SUNMemory files directly, so I am updating them in place. These latest files are necessary for DPC++.

Other notes about why this is the case for anyone interested:
1. For CMake, we choose not to enable SUNDIALS through AMReX as intended because it requires us to build SUNDIALS prior to Pele and point AMReX to a SUNDIALS installation, which is much less convenient for us than to just `add_subdirectory()` to the SUNDIALS submodule and build SUNDIALS along with Pele.
2. For GNUmake, I was not able to point to AMReX's `Src/Extern/SUNDIALS/Make.package` correctly so that linking to the SUNMemory objects would succeed and we could include the files from AMReX directly instead of storing copies here. It links fine when the files are local to `PelePhysics/Reactions`. If GNUmake could point directly to AMReX's files and succeed, I would prefer that, and I'm pretty sure I could get that to work in CMake, but until then I think we have to store copies of them here.


marchdf (approved):

esclapez:
Do you want to try have GNUmake point at AMReX's ?

jrood-nrel:
Yes, if you can make it work, that would be great. I tried:
```
Bpack += $(AMREX_HOME)/Src/Extern/SUNDIALS/Make.package
Blocs += $(AMREX_HOME)/Src/Extern/SUNDIALS
```
but it did not work because the objects didn't show up on the link line. I also tried several other things, but had to give up.

esclapez:
I just tried the Bpack/Blocs option and it seems to work to compile ReactEval. Are you having issue compiling PeleC ?

jrood-nrel:
Interesting. Yeah if you look at my first commit, the AMReX_SUNMemory.cpp object file wouldn't show up on the PeleC link line. I spent like an hour trying to understand why. It just seems like it needs a relative path instead of absolute for some reason.

esclapez:
Checking compiling LM now. It looks like in your first commit you were adding to `CEXE`s directly, but you didn't have the Blocs. You probably tried that but not commited/pushed, though. Not sure what I'm doing differently. I'm on Eagle, with the environment you provided me a little while ago.

jrood-nrel:
Thanks for testing. I will give it another try because it really should work.

esclapez:
I have the same issue as you when compiling LM: error at linking. There must be something we do differently than ReactEval.

jrood-nrel:
Ah, good to know I'm not the only one who sees that.

esclapez:
Pretty odd: `make print-Bpack` does list the Extern/SUNDIALS, but then `make print-CEXE_headers` doesn't list the content of that folder. There must be an overwrite somewhere.

esclapez:
So, not particularly clever, but:
```
 ifneq (, $(filter TRUE, $(USE_CUDA) $(USE_HIP) $(USE_DPCPP)))
     Blocs += $(AMREX_HOME)/Src/Extern/SUNDIALS/
     CEXE_sources += AMReX_SUNMemory.cpp
     CEXE_headers += AMReX_SUNMemory.H
     CEXE_headers += AMReX_Sundials.H
     VPATH_LOCATIONS += $(AMREX_HOME)/Src/Extern/SUNDIALS                                                                                                                     
     INCLUDE_LOCATIONS += $(AMREX_HOME)/Src/Extern/SUNDIALS
 endif
```
seems to work. I still can't find why the plain `Bpack` works in PelePhysics but not in C/LM.

jrood-nrel:
Awesome. That worked for me. The last thing to note when we merge this is you will have to change the `Initialize` calls in LM, but obviously that's apparent in the reacteval changes.

esclapez (approved):

PR #209, opened 2021-10-14
Updates to remove reliance on managed memory

marchdf:


jrood-nrel (approved):

jrood-nrel:
I think we should merge this.

jrood-nrel:
I guess what I mean is PeleLM doesn't need managed memory gone to use this right?

PR #210, opened 2021-10-18
Fix OMP data races and std output in Reactions

jrood-nrel:


jrood-nrel:
This solves all data races I can find with the thread sanitizer for rk64/arkode/cvode. I don't know how cvode ever worked with openmp. This is a temporary fix for cvode to make it at least correct in openmp because I have to use one large `#pragma omp single`. This relies on this branch as well https://github.com/AMReX-Combustion/PelePhysics/pull/209 .

esclapez:
Cvode probably wasn't working with OMP, I haven't tested it since the runtime selection was implemented but there seems to be some issue https://github.com/AMReX-Combustion/PeleLM/issues/199. I'll try with this branch to see if it fixes the problem.

marchdf (approved):

PR #214, opened 2021-10-28
Update PMF handler

esclapez:
- first pass at providing a straightened framework to handle PMF data and host/device locations.
 - new Testing/Exec/PMFReader tested on CPU & GPU CUDA w & w/o managed memory.


drummerdoc (approved):

Testing/Exec/Make.PelePhysics
```diff
@@ -132,7 +136,7 @@ endif
 
 include $(AMREX_HOME)/Tools/GNUMake/Make.rules
 
-$(objForExecs): TPL
+#$(objForExecs): TPL
```
marchdf: Is this intended?

Testing/Exec/PMFReader/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+main(int argc, char* argv[])
+{
+  Initialize(argc, argv);
+
+  {
+    Real strt_time = amrex::ParallelDescriptor::second();
+
+    BL_PROFILE_VAR("main::main()", pmain);
+
+    // -----------------------------------------------------------------------------
```
marchdf: not a fan of these kinds of comments ;) I would be in favor of removing the `--------` lines.

Testing/Exec/PMFReader/main.cpp
```diff
@@ -0,0 +1,131 @@
...
+
+    Vector<int> n_cell(AMREX_SPACEDIM);
+    pp.getarr("n_cell",n_cell,0,AMREX_SPACEDIM);
+
+    IntVect lo(IntVect::TheZeroVector()), hi(n_cell);
+    hi -= IntVect::TheUnitVector();
+    Box domain(lo,hi);
+
+    amrex::RealBox *real_box = nullptr;
+   int coord = -1;
```
marchdf: Indentation? Feed this through `clang-format`?

Testing/Exec/ReactEval/extern_parameters_F.H
```diff
@@ -0,0 +1,17 @@
+
```
marchdf: Do we need this file?

Utility/PMF/PMFData.H
```diff
@@ -0,0 +1,98 @@
...
+      unsigned int m_nVar;
+      int m_doAverage = 0;
+      amrex::Real* pmf_X;
+      amrex::Real* pmf_Y;
+  };
+
+  void allocate ()
+  {
+    if (!m_device_allocated) {
+        m_data_d = (DataContainer*) amrex::The_Device_Arena()->alloc(sizeof(m_data_h));
```
marchdf: I think we should make this `The_Arena` instead of the Device_Arena. Should be done in the transport as well. I know they are the same but I think it is clearer.

marchdf (changes requested):

Testing/Exec/PMFReader/main.cpp
esclapez: I like those, but I can remove it here

Testing/Exec/ReactEval/extern_parameters_F.H
esclapez: No, I don't think we need extern at all anymore. Ill try to remove it from the GNUmake actually.

Utility/PMF/PMFData.H
esclapez: I agree

Testing/Exec/PMFReader/main.cpp
marchdf: haha again, totally a preference thing. Keep if you want ;)

marchdf (approved):

jrood-nrel (approved):

PR #216, opened 2021-11-02
Fix to some PMF warnings

marchdf:


marchdf:
@esclapez the only commit to look at here is  the `Fix to some PMF warnings` commit (first one of this PR). The rest is just clang-format doing its thing. Basically the PMF stuff triggered some warnings in PeleC so I tried to fix them.

esclapez (approved):
Looks good !

PR #218, opened 2021-11-02
static_asserts for transport and eos models

baperry2:
Fixes error checking for when invalid transport or EOS models are specified.


marchdf:
Yikes! good catch!

marchdf (approved):

PR #219, opened 2021-11-02
Fix PMF free()

esclapez:
Only free() the host data if it has been allocated.

### DIFF
--- a/Utility/PMF/PMFData.H
+++ b/Utility/PMF/PMFData.H
@@ -55,10 +55,13 @@ public:
 
   void deallocate()
   {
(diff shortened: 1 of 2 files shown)


marchdf (approved):

marchdf:
I don't think this is the end of the story... Jon tells me something is wrong when not using managed memory. And it seems slower: https://github.com/AMReX-Combustion/PeleC/pull/370. Any thoughts?

esclapez:
I tested it in the Testing/Exec/PMFReader and it seemed to work, but I haven't checked in LM or LMeX yet. I'll setup some LM test once this is merged.

PR #220, opened 2021-11-08
Fix Analytical Jacobian options on GPU.

esclapez:
CVODEUserdata need to allocated with The_Arena when working on GPU otherwise the analytical jacobian data are not accessible.


marchdf:
yes but... will this work without managed memory? I think the assumption I was operating under is that this is a pure host structure so we don't need a device copy. Obviously I am wrong with this assumption for the analytical jacobian stuff.

ldowen:
Whatever happened to this fix? was it taken care of separately?

esclapez:
I'll close this PR and delete that branch soon since this problem, and other, are addressed in #233

PR #222, opened 2021-11-12
Use parallel build for magma and sundials

jrood-nrel:
I wasn't getting parallel builds for magma, but this is building much faster for me. I also merged in `fix_AJ_on_GPU`.


esclapez (approved):
I don't know enough about cmake :) But it does go a lot faster indeed. Thanks !

PR #223, opened 2021-12-01
clip C before reaction rate calculations

baperry2:
Opening this for discussion in conjunction with Marc's PeleC pull requests https://github.com/AMReX-Combustion/PeleC/pull/381 and https://github.com/AMReX-Combustion/PeleC/pull/380

Don't allow negative C when computing WDOT. This is similar in spirit to the "clean_massfrac" function in PeleC, but applies only to reaction rates and only clips negatives. clean_massfrac seems to help things often, but can create issues of its own - perhaps due to rescaling. In contrast, I think this is something that can be safely left on all the time.

Say you have a reaction `A + A -> products` and A is negative, or `A + B -> products` and A,B are both small negative values for whatever reason. Then the consumption rate of A is positive (k_f*negative*negative), so A will get smaller - making it more negative, and leading to a runaway. This change prevents that - once A became negative it's consumption rate would always be 0, preventing runaway. A downside is that for unimolecular consumption of A, allowing negative mass fractions for reaction rate computation has a stabilizing effect (A would have a negative consumption rate, aka production, driving it toward 0/positivity).

### DIFF
--- a/Eos/Fuego.H
+++ b/Eos/Fuego.H
@@ -168,7 +168,7 @@ struct Fuego
 
     // Do not allow negative C for wdot calculation
     for (int ii = 0; ii < NUM_SPECIES; ++ii) {
-      C[ii] = amrex::max(C[ii],0.0);
+      C[ii] = amrex::max(C[ii], 0.0);
     }
 
     CKWC(&T, C, WDOT);
--- a/Eos/Fuego.H
+++ b/Eos/Fuego.H
@@ -165,6 +165,12 @@ struct Fuego
     // CKWYR(&R, &T, Y, WDOT);
     amrex::Real C[NUM_SPECIES];
     CKYTCR(&R, &T, Y, C);
+
+    // Do not allow negative C for wdot calculation
+    for (int ii = 0; ii < NUM_SPECIES; ++ii) {
+      C[ii] = amrex::max(C[ii],0.0);
+    }
+
     CKWC(&T, C, WDOT);
 
     amrex::Real mw[NUM_SPECIES];
--- a/Eos/SRK.H
+++ b/Eos/SRK.H
@@ -555,6 +555,12 @@ struct SRK
   {
     amrex::Real C[NUM_SPECIES];
     RTY2C(R, T, Y, C);
+
+    // Do not allow negative C for wdot calculation
+    for (int ii = 0; ii < NUM_SPECIES; ++ii) {
+      C[ii] = amrex::max(C[ii],0.0);
+    }
+
     CKWC(&T, C, WDOT);
 
     amrex::Real mw[NUM_SPECIES];
--- a/Eos/SRK.H
+++ b/Eos/SRK.H
@@ -558,7 +558,7 @@ struct SRK
 
     // Do not allow negative C for wdot calculation
     for (int ii = 0; ii < NUM_SPECIES; ++ii) {
-      C[ii] = amrex::max(C[ii],0.0);
+      C[ii] = amrex::max(C[ii], 0.0);
     }
 
     CKWC(&T, C, WDOT);


drummerdoc:
For the reaction O+O<=>O2, negative O drives more negative O.  I have tried clipping C prior to WDOT eval, and that helps, but problems can still exist if the A+D forcing wants to drive the O negative anyway.

marchdf (approved):

marchdf:
I am hoping that the EB SRD problem/fix might make this less necessary... I still think it's a good idea.

esclapez (approved):
That'll definitely make things more robust.

ldowen:
I gave this PR a test on a reacting spray case and it seemed to cause my time steps to take a very long time. Like 10x longer. I'm not 100% sure why but I wonder if it is due to a disconnect between the concentrations used in computing omega dot and the concentrations that are being updated. Has anyone else seen this sort of issue?

marchdf:
I think @olgadoronina saw something like this in her case when using this. @drummerdoc have you seen this before when using this fix?

drummerdoc:
Yes.  I mentioned above one cause of additional problems... that the auxiliary forcing continues to drive the concentrations negative.  there is also that there are other EOS calls in the chemistry RHS eval that are sensitive to negative mass fractions and those aren't being clipped (as suggested by Landon).  When faced with this in the past I tried to then climb into CVODE and prevent the generation of negatives before they happen by futzing with the state updates, and that turns out to be extremely icky because of how the updates and time stepping are managed internally.  

This does remind me though that some time ago, Carol had a student add "constraints" to the updates for exactly this reason, but I know that we are not exercising that bit, and I'm not sure if the constraint stuff made it into the GPU versions of SUNDIALS. It might be worth asking.

For PeleLM, back in the day, this became a significant enough of a mess, that (as I mentioned in a recent call), we actually took a more brute force approach to fixing it.  We detected that a mass fraction became sufficiently negative during the evolution that things were destined to fail (since we applied the JBB_HACK to ensure we started in a reasonable place).  If that happened, we kicked out of the integration entirely, and restarted the integration in that cell with the auxiliary forcing zeroed.  After the cell integrated successfully, we would add the Faux.dt term to the result directly for consistency, even though it drove the result horribly negative. (if it still failed, we aborted - something was terribly wrong)

But, since the GPU version works on all cells simultaneously, that approach cannot be applied in a straightforward way and requires more thought.

All that said, the first step is to truly understand why the states are going negative.  It is either that the tolerances are not sufficiently tight to keep the evolution well behaved, or the auxiliary forcing is driving things out.  Once we understand that bit, we can try to engineer a more robust scheme.  

Finally, if the code really wants to step out negative due to a strongly negative chemical source term, for example, it will cut the internal time step very small to avoid that and will end up taking forever to integrate.  In my experience, the only remedy is to understanding the typVals/scaling stuff in gory detail to see if you can dramatically improve the integration by scaling some of the components more correctly.

IMG_1413

drummerdoc:
In the pic above, if the scaling is set correctly, CVODE will find a numerical dt that is reasonable and can make progress.  If the scaling is completely out of whack, it could end up over correcting and cutting dt to something extremely small for safety and then take many many steps to jack the step back up to get out of that hole.

hsitaram:
pelec now has a feature to set typvals. use flags pelec.use_typ_vals_chem=1
ode.rtol=1e-4 ode.atol=1e-5. It can help with CVODE convergence.

drummerdoc:
Sorry for the spam...

In the past (on CPU), I have used std::cout to find the problematic cell, then grab all the inputs (initial state, post JBB_HACK state, dt_cfd, auxiliary forcing, typVals) and reproduce the failure within PelePhysics' ReactEval driver code, for example.  Once there, I added std::cout's in the chem RHS function and dumped time and the state until it fails, then plot the trajectory of the bad actors and use that to dig into what was happening.  (in the end, it always comes to dumping numbers to the screen).  But then, you can adjust typVals, for example, and tolerance scalars, shut off WDOT, and other tricks to nail down precisely what is happening.  Then again, I'm old school. There's probably a clever Cantera or python way of doing this point-and-click or something :)

ldowen:
The tricky thing is defining "failure." For the test case I am running, we get unphysical temperature spikes that show up over time. Turns out these regions of high temperature correspond with large negative mass fractions of intermediate species (around -0.0008 I think). The solvers take much longer to fail but the problem is already present.

ldowen:
Also, this is a spray test case, so we can't rule out the possibility that the added continuity source terms from evaporation are also causing issues.

marchdf:
In PeleC, I believe these negative mass fractions come from EB and how we are computing fluxes. I am still digging into this. For PeleLM + a spray case, a careful investigation of why these are happening has to happen. Doing this for PeleC enabled me to find two real fixes (not hacks, actually incorrect things). There still remains the question of EB... 

So I have nothing to suggest but the long and painful way... 

I also think we should try implementing the JBB_HACK in pelephysics...

esclapez:
I agree on the need for testing JBB_hack. Nick and I are looking into LMeX species overshoots and it seems that PPM overshoot massively (e-2) while PLM is better behaved.

drummerdoc:
Yes, if nothing else the JBB_HACK guarantees that you start with a reasonable state.  If it goes negative during the integration, you'll have something to look for.

marchdf:
@esclapez how are the slopes computed? Can you ensure that `rho` at the face is `= sum rhoY`? I think this part is crucial.

esclapez:
> @esclapez how are the slopes computed? Can you ensure that `rho` at the face is `= sum rhoY`? I think this part is crucial.

We reconstruct the rho at the faces from the `rhoYs`. The overshoot/undershoot are consistent in that rho remains equal to the sum of `rhoYs`.

Eos/Fuego.H
```diff
@@ -165,6 +165,12 @@ struct Fuego
     // CKWYR(&R, &T, Y, WDOT);
     amrex::Real C[NUM_SPECIES];
     CKYTCR(&R, &T, Y, C);
+
+    // Do not allow negative C for wdot calculation
+    for (int ii = 0; ii < NUM_SPECIES; ++ii) {
```
hsitaram, line 170: I understand this fix is in the lower level fuego routines but can there be flag to turn this on and off by the user?

baperry2: Based on the above discussion showing slowdowns with this change,  I agree making this a runtime option would be better. But with the way the EOS code is presently set up, that would be hard to do. There's no way to pass input file parameters to EOS calls right now (e.g. even for GammaLaw EOS, Gamma is hardcoded).  

So I don't think this PR should be integrated, at least for now. It may be a component of robustification, but I think it needs to wait until we better understand what's going on.

drummerdoc:
Since ParmParse queries are cheap, a quick and dirty way to do this would be to ParmParse the option into either the base reactor class (preferred) or into each of the reactors to set member data that can be pushed onto the device like prob data.  I'm guessing that when you figure out why the flooring slows things down, it will lead to a couple more options :)

marchdf:
@baperry2 should we close this for now?

baperry2:
Yeah, I think it is basically superseded by JBBhack

PR #226, opened 2021-12-28
Update to SUNDIALS 6

balos1:
Needs https://github.com/AMReX-Codes/amrex/pull/2551. @jrood-nrel @esclapez @gardner48


esclapez:
@balos1 Hi Cody. Currently, some clang formatting issues are blocking this PR. Can you address those ? Alternatively I can do it if you give me the right to push to your branch. Thanks !

jrood-nrel:
I just verified the CI checked out the `1404c4a` version of AMReX. I'm not sure why that header isn't being found but I haven't looked very hard. @balos1 Do you want any help solving this or do you have an idea what's going wrong?

gardner48:
@jrood-nrel I ran into this issue with PeleLMeX and the problem was `amrex/Src/Extern/SUNDIALS` not getting included. Adding `USE_SUNDIALS=TRUE` to the Makefile fixed the issue for me.

esclapez:
We'll always need SUNDIALS now, so we can probably add this to `Testing/Exec/Make.PelePhysics` and the equivalent in C and LM. But maybe this also triggers other things.

jrood-nrel:
Yeah I think we kind of do all the sundials stuff manually. Maybe https://github.com/AMReX-Combustion/PelePhysics/blob/development/ThirdParty/Make.ThirdParty could be less complicated, as well as https://github.com/AMReX-Combustion/PelePhysics/blob/development/Reactions/Make.package

jrood-nrel:
It looks like `Make.ThirdParty` does get simplified here. It looks like transport doesn't have `USE_SUNDIALS` but the other tests do.

balos1:
> We'll always need SUNDIALS now, so we can probably add this to `Testing/Exec/Make.PelePhysics` and the equivalent in C and LM. But maybe this also triggers other things.

Yes, this is what needs to be done and what I did in PeleC.

balos1:
@esclapez Looks like all the checks are passing now.

esclapez:
Awesome. I'll give a try to LM on both CPU/GPU with this version and it should be good to go on my side.

esclapez:
All good on LM/LMeX side.

jrood-nrel (approved):

PR #232, opened 2022-02-25
Refactor turbulence inflow

marchdf:
This simplifies the turbulence inflow, encapsulates it better, makes it more modular. The user will have to switch to a new PP querry prefix (`turbinflow` instead of `prob`).

@esclapez can you give this a shot on a problem that works for you in LM? You can see what I did to PeleC in the PR referred below.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: DoozyX/clang-format-lint-action@v0.11
         with:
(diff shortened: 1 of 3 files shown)


marchdf:
Mohammad tested this out and it seems to work. I think @esclapez is doing the same on the LM side. 

Screen Shot 2022-02-27 at 12 11 25 PM

esclapez (approved):
Tested in both LM and LMeX, on CPU & Nvidia GPU successfully.

PR #234, opened 2022-03-09
Add a PltFileManager

esclapez:
- handling reading data out of the file
 - managing fillPatch operations to fill provided grid patches (zero ghost cell only !)
 
This can be used in place of the initdata function to provide initial data at a given level.

### DIFF
--- a/Utility/PltFileManager/PltFileManager.H
+++ b/Utility/PltFileManager/PltFileManager.H
@@ -7,12 +7,13 @@
 
 namespace pele {
 namespace physics {
+namespace pltfilemanager {
 
 class PltFileManager 
 {
 public:
 
-  void init(const std::string &a_pltFile);
+  PltFileManager(const std::string &a_pltFile);
 
   void readGenericPlotfileHeader(const std::string &a_pltFileHeader);
 
@@ -41,6 +42,7 @@ protected:
   amrex::Vector<amrex::MultiFab> m_data;
 };
 
+} // namespace pltfilemanager
 } // namespace physics
 } // namespace pele
 #endif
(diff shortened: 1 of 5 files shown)


Utility/PltFileManager/PltFileManager.H
```diff
@@ -0,0 +1,46 @@
+#ifndef _PLTFILEMANAGER_H_
+#define _PLTFILEMANAGER_H_
+
+#include <AMReX_MultiFab.H>
+#include <AMReX_ParmParse.H>
+#include <AMReX_Geometry.H>
+
+namespace pele {
+namespace physics {
+
```
marchdf, line 11: Would you mind namespacing one more time? `namespace pltfilemanager` That way it follows the turbinflow nesting and the other nestings as well.

Utility/PltFileManager/PltFileManager.H
```diff
@@ -0,0 +1,46 @@
...
+class PltFileManager 
+{
+public:
+
+  void init(const std::string &a_pltFile);
```
marchdf: can we just have a constructor instead of an `init` ? Or is this because we are using it as a `static` in LM/C?

Utility/PltFileManager/PltFileManagerBCFill.H
```diff
@@ -0,0 +1,22 @@
+#include <AMReX_PhysBCFunct.H>
+#include <AMReX_Geometry.H>
+
```
marchdf, line 3: maybe namespace this as well into `pele::physics::pltfilemanager` ?

marchdf:
This is awesome! Just a couple minor thoughts.

marchdf (approved):

PR #238, opened 2022-03-14
Use modern loops in Reactor

jrood-nrel:


esclapez:
@jrood-nrel does this type of change provides performances gain ? Or this is more a style change ?

jrood-nrel:
It's for a few reasons. First, clang-tidy prefers it. Second, it allows us to remove `NUM_SPECIES`. It looks like these loops work in CUDA. I think they work on GPUs in general, but I'm researching it. It's not for performance.

esclapez (approved):
Can we use `amrex::Real&` instead of `double&` ?
Other than that, if it works with GPU compilers that's neat.

jrood-nrel:
Ah yes. Good catch.

marchdf (approved):

Reactions/ReactorUtils.H
```diff
@@ -104,8 +104,8 @@ box_flatten(
     mass_frac[n] = rhoYn;
   }
   const amrex::Real rho_inv = 1.0 / rho;
-  for (int n = 0; n < NUM_SPECIES; n++) {
-    mass_frac[n] *= rho_inv;
+  for (amrex::Real& n : mass_frac) {
+    n *= rho_inv;
```
marchdf: this is stupid but since we use `n` to index things, can we use `mf` instead of `n` here? That way we know that we are *= a massfrac?

drummerdoc: ...or Yn

marchdf: Yn is even better lol

PR #239, opened 2022-03-16
Dodecane methane air mix

esclapez:
Add a mixture of dodecane/methane/air mechanism for testing mixing of the KPP challenge problem.


marchdf (approved):

esclapez:
@jrood-nrel just intrigued as to why the CI isn't popping up. Did I break something or somehow it catches that I'm only adding things and not touching any of the code that is actually tested ?

jrood-nrel:
It's down across github.

PR #240, opened 2022-03-28
Multiple TurbParm

esclapez:
It is now possible to use multiple TurbParm.
One needs to specify the number of TurbInflows:
`turbinflow.num_inflows = 2`
Then each TurbInflow properties is specified using `turbinflow1`, `turbinflow2`, ...
Additionally, one now to provide on which direction and side each turbulence dataset is expected to act:
`turbinflow1.dir            = 1`   (0, 1 or 2 for x, y and z, resp.)
`turbinflow1.side          = "low"` (low or high)

A bug fix in the z-dir indexing during interpolation is also implemented.

### DIFF
--- a/Support/Fuego/Pythia/pyproject.toml
+++ b/Support/Fuego/Pythia/pyproject.toml
@@ -12,7 +12,7 @@ scikit-learn = "^0.24.2"
 future = "^0.18.2"
 
 [tool.poetry.dev-dependencies]
-black = {version = "^21.7b0", allow-prereleases = true}
+black = {version = "^22.3"}
 isort = "^5.9.3"
 
 [build-system]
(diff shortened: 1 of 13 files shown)


drummerdoc:
Just an idea for implementation: what if the turbinflows were named:

`turb_inflows = bottom_jet lateral_jet`
` turb_inflow.bottom_jet.dir = 2`
`turb_inflow.bottom_jet.side = low`
`turb_inflow.lateral_jet.dir = 1`
`turb_inflow.lateral_jet.side = high`

esclapez:
That should work too, probably better.

marchdf (approved):

marchdf:
This is excellent. Thanks for doing this!

marchdf:
Oh, do we want to add an "offset time" thing? Maybe we can add a time offset that defaults to zero and can be switched to random? Or we init it to something random. The trick to this is that it should work on restart... and that is non-trivial even if setting a seed. So maybe it's best to have the user input the time offset they want for each turbinflow? Just thinking out loud here...

drummerdoc:
I think the time offset will be critical to avoid "beating" phenomena from multiple jets.

esclapez:
It is in there already: `time_shift` in the TurbParm struct, parsed as `time_offset`.
It shifts the calculation of `z` : `(time+tp[n].time_shift) *`...

PR #242, opened 2022-03-31
Add SUNDIALS accelerated fixed-point solver option

gardner48:


jrood-nrel (approved):

esclapez:
Hi @gardner48,

I gave it a try on CPU and I think it needs a couple of `if (LS != nullptr)` to prevent accessing the linear solver which do not exist:
 - when calling `CVodeSetJacEvalFrequency`
 - when calling `CVodeGetNumLinRhsEvals`

Except those, it works nicely. I'll try to play with it and see if and where this can be used in place of the Newton.

gardner48:
Hi @esclapez, I think I've addressed all the other places where a `if (LS != nullptr)` check was needed. Let me know if you spot any other issues.

esclapez (approved):

Issue #248, opened 2022-04-14
CPickler error while trying to generate mechanism files from CHEMKIN


RSuryaNarayan:
Hello, 

I am trying to parse a high-pressure DME mechanism [here](https://ars.els-cdn.com/content/image/1-s2.0-S0010218018302396-mmc2.zip) into Pele and run into this error after doing ```./make_mechanism.sh```. 
```
Traceback (most recent call last):
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/products/bin/fmc.py", line 22, in <module>
    app.main()
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/pyre/pyre/applications/Application.py", line 28, in main
    self.execute(*args, **kwds)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/pyre/pyre/applications/Application.py", line 45, in execute
    self.run(*args, **kwds)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/fuego/fuego/fmc/FMC.py", line 59, in run
    lines = fuego.serialization.save(mechanism, output)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/fuego/fuego/serialization/pickle.py", line 33, in save
    return pickler.pickle(mechanism)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/weaver/weaver/mills/Mill.py", line 26, in pickle
    self._renderDocument(document)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/fuego/fuego/serialization/c/CPickler.py", line 302, in _renderDocument
    self._thermo(mechanism)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/fuego/fuego/serialization/c/CPickler.py", line 745, in _thermo
    speciesInfo = self._analyzeThermodynamics(mechanism, 0)
  File "/home/suryanarayan/Desktop/PelePhysics/Support/Fuego/Pythia/pythia-0.5/packages/fuego/fuego/serialization/c/CPickler.py", line 716, in _analyzeThermodynamics
[...]

PR #249, opened 2022-04-14
Adding HP_DME

marchdf:
This doesn't work yet but I wanted to make sure we are all using the same files to debug this issue. @malihass had some suggestions and I've only done some of them. So for a list of changes that I've made:
- [x] found and removed another `#4`
- [x] fixed whitespace issues in `therm.dat`
- [x] added the `TRAN` data into `mechanism.inp`
- [x] there are duplicates in `therm.dat` we need to remove them 
- [x] renaming O*(SING) -> O*SING, that way we don't lose those reactions, and then catching those in the Cpickler and using O weight
- [x] I fixed the pickler so it ignores PLOG and throws a warning instead of erroring out. So we keep those reactions, just don't use PLOG
- [x] There's a TROE line that was using commas, removed those commas

This relates to #248


hsitaram (approved):
nicely done!

drummerdoc (approved):
Looks great!

malihass (approved):

PR #250, opened 2022-04-14
Fix NULL use, some const params, and don't use implicit bools

jrood-nrel:


esclapez:
For my education: implicit bools are bad or just bad style ? I use them all the time ...

esclapez (approved):

jrood-nrel:
I think it's both. Completely overt code can help to reduce the possibility of unintended behavior. Letting the compiler handle things implicitly place the burden on the programmer to know the default behavior of something. The better fix would be to switch a lot of these to bools to be even more clear. These are mostly minor though. I'm really just cleaning out the noise to hopefully catch any real bugs popping up in the future.

PR #255, opened 2022-04-18
Clang-format mechanisms

marchdf:


jrood-nrel (approved):

malihass:
Hi, 
It seems that dodecane_lu_qss is not in the list of mechanisms. Also the mechanism.H and mechanism.cpp have disappeared from the dodecane_lu_qss. I think this is because the QSS tools use `future`  and `numpy` (not the case of the other mechanism) and if we just run `poetry run` without doing `poetry install` the necessary packages are not installed. @jrood-nrel @marchdf How should we fix it?

marchdf:
I didn't include `dodecane_lu_qss` because I wasn't sure it would regenerate out of the box. Maybe create a PR that adds it to that list and make sure it regenerates ok? We don't run (I think) the regeneration script through poetry so I don't think that's the problem?

malihass:
Just for reference: this is the object of PR #268

PR #260, opened 2022-04-20
Ignition delay test case

malihass:
Adding a test case for computing the ignition delay of a mechanism.
This uses a lean version of ReactEval where function for initialization and reactions are separated from the main script. The ignition delay is computed in 2 stages: first it is roughly estimated; second that estimate is used to set up a finer calculation of the ignition delay.
A python script is in charge of processing the result of the 0D reaction and setup the second stage of the ignition delay calculation.


Testing/Exec/IgnitionDelay/computeIgnitionDelay.py
```diff
@@ -0,0 +1,80 @@
...
+                key, value = splitLine
+            else:
+                key, value = splitLine[:2]
+            inpt[key.strip()] = value.strip()
+    f.close()
+
+    return inpt
+
+def writeNewInput(oldInput,newInput,dt,ndt):
+    f = open(oldInput,'r+')
```
marchdf: this kind of thing (here and other spots) is best done with a context manager:
```
with open(oldInput,'r+') as f:
    lines = f.readlines()
```

Testing/Exec/IgnitionDelay/computeIgnitionDelay.py
```diff
@@ -0,0 +1,80 @@
...
+        elif line.startswith('ode.ndt'):
+            f.write('ode.ndt = %d ' % (ndt))
+            f.write('\n')
+        else:
+            f.write(line)
+    f.close()
+
+    return
+
+# CLI
```
marchdf: can we protect this with `if __name__ == "__main__":`

Testing/Exec/IgnitionDelay/utils/fileio.py
```diff
@@ -0,0 +1,261 @@
+import numpy as np
```
marchdf: do we need this? feels like we can just use pandas or something?

Testing/Exec/IgnitionDelay/hrTest.sh
```diff
@@ -0,0 +1,25 @@
+# Find first executable name available
```
marchdf: it's unfortunate that we need these script files for these tests. @jrood-nrel do you have an idea for what we could do here? Is this best done within `ctest`?

Testing/Exec/IgnitionDelay/computeIgnitionDelay.py
```diff
@@ -0,0 +1,80 @@
...
+
+# Process result
+filePP = 'PPreaction.txt'
+A1 = io.readMultiColFile(filePP)
+ignitionDelayPP = (dt/ndt)*np.amin(np.argwhere(A1[:,1]>(A1[0,1]+400)))
+
+
+if args.estimate:
+    if args.verbose: 
+        print("Rough estimate: T = %.3g K , P = %.3g atm, Phi = %.3g,  tIgn = %.3g +/- %.3g s " % (temp,press/101325.0,phi,ignitionDelayPP,dt/ndt))
```
marchdf: I would encourage the use of `fstrings` instead of this old style print formatting

Testing/Exec/IgnitionDelay/utils/fileio.py
malihass: Is there a memory concern that would encourage us to use pandas over numpy? 
I think not using numpy might make the code a bit more messy, no?

Testing/Exec/IgnitionDelay/hrTest.sh
jrood-nrel: I think it's fine. Adding ctest would be a lot more work. Just add `#!/bin/bash` here.

jrood-nrel (approved):

jrood-nrel:
Do we want to have this run on Github?

Testing/Exec/IgnitionDelay/hrTest.sh
malihass: Fixed the header, but I leave this unresolved in case @marchdf wants to try another method

Testing/Exec/IgnitionDelay/utils/fileio.py
marchdf: Oh sorry, I didn't mean numpy vs pandas. I meant the whole `fileio` thing.

marchdf (approved):

marchdf:
Looks good! Thank you for doing this!

malihass:
It might be a good idea to have this running on Github. Just checking the test runs is a good indication that nothing too weird is going on. Ideally, in the future, we will want to be able to compare the output of the ignition delay test with a theoretical value.

esclapez (approved):

Testing/Exec/IgnitionDelay/.githubTest.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
```
marchdf: @jrood-nrel if this is executed by the CI, should it be somewhere else?

Testing/Exec/IgnitionDelay/computeIgnitionDelay.py
```diff
@@ -0,0 +1,108 @@
...
+            if line.startswith('ode.dt'):
+                f.write('ode.dt = %.10f' % (dt))
+                f.write('\n')
+            elif line.startswith('ode.ndt'):
+                f.write('ode.ndt = %d ' % (ndt))
+                f.write('\n')
+            else:
+                f.write(line)
+
+    return
```
marchdf: remove all empty "return"

Testing/Exec/IgnitionDelay/exec_ignDelay.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
```
marchdf, line 1: This looks like it's the same as the other sh script. Should we somehow consolidate?

marchdf (approved):

Testing/Exec/IgnitionDelay/.githubTest.sh
jrood-nrel: I think it's better here. I don't think it should be a hidden file though.

Testing/Exec/IgnitionDelay/exec_ignDelay.sh
malihass: So the difference was that the `exec_ignDelay.sh` is the file that a user would use to compute ignition delay. The `githubTest.sh` was meant to do it AND check that the value obtained was reasonable. This was done by passing an argument to the `computeIgnitionDelay.py` and signify that a GitHub action is done or no.  I also think this was a bit redundant, and that the test should actually not be buried in the ignition delay computation.

I got rid of the two bash files and only use `exec_ignDelay.sh`. Now `computeIgnitionDelay.py` is not in charge of checking that the value is ok anymore. It outputs the value to a log file which is read and checked during the github action.

marchdf (approved):

PR #262, opened 2022-04-20
Use absolute rpath for TPL

jrood-nrel:
Barely know what I'm doing in GNUmake, but mostly trying to get the rpath to use the absolute path to SUNDIALS.

### DIFF
--- a/ThirdParty/GNUmakefile
+++ b/ThirdParty/GNUmakefile
@@ -49,7 +49,7 @@ else
   endif
 endif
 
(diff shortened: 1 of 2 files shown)


jrood-nrel:
This also tries to solve the case of `USE_CUDA=TRUE COMP=gnu` which didn't previously work.

marchdf (approved):

jrood-nrel:
I've been using this a lot today and it appears to work how I expect. Going to merge it and fix stuff if we have to after.

PR #266, opened 2022-04-28
Switch to sundials 6.2

jrood-nrel:


esclapez:
Can we do magma 2.6.2 at the same time ?

esclapez (approved):

PR #267, opened 2022-04-29
Replace device vector with host vector in turbinflow

jrood-nrel:


esclapez:
I'll give it a whirl in LMeX on Summit (Crusher's down ...)

jrood-nrel:
So far I've only verified it doesn't segfault.

esclapez (approved):
It seems to have resolved the issue, thanks !

Issue #271, opened 2022-05-10
TypeError: list indices must be integers, not str


czc-zju:
To build a new CHEMKIN based model, the TRANFILE_APPEND.txt is generated and inserted into the chemistry input file,  but when I run the script file make-mechanism.sh, TypeError: list indices must be integers, not str.

marchdf:
We need more specifics... what is the chemkin file(s)? where is python reporting the error?

malihass:
Hi, 
thanks for submitting these files. 
It looks like this is an issue with R16: `H2O2(+H2O)<=>OH+OH(+H2O)`
When we compute the third body mixture, if `(+H2O)` was `(+M)` we would just compute the mixture value. Here, we need to read `sc[H2O.id]`. It seems that the `H2O.id` is not correctly obtained. I will submit a PR to fix this. 
I tested it after the code fix and it seems to work. I encourage you to try it once the PR goes through.

Attached are the mechanism.cpp and mechanism.H 
[mechanism.H.txt](https://github.com/AMReX-Combustion/PelePhysics/files/8663046/mechanism.H.txt)
[mechanism.cpp.txt](https://github.com/AMReX-Combustion/PelePhysics/files/8663048/mechanism.cpp.txt)


Let me know if this works and I will close this

czc-zju:
It does work, it's amazing!!!!!!!!
Thanks a million.
That's awesome
How do you know which reaction is causing the problem?

malihass:
No prob. Closing this then.

PR #272, opened 2022-05-10
Fix indexing of third body species

malihass:
This is related to Issue #271 
Third body species were not correctly indexed.
You can verify this by running the vanilla and modified `CPickler.py` on the `hydrogen_cyanide` case


Support/Fuego/Mechanism/Models/hydrogen_cyanide/README.md
```diff
@@ -0,0 +1,8 @@
+# Analytically Reduced Mechanisms
```
marchdf: do we want this file?

Support/Fuego/Mechanism/Models/hydrogen_cyanide/make-mechanism.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
```
marchdf: can we make this `make-mechanism.sh` look like the others? And then just move the *.inp and *.dat into this folder (instead of the `orig` folder)?

malihass: I originally did not plan to keep that mechanism here since we did not verify it (it comes from the issue that was opened). Should we keep that in the repo? if yes I can standardize it and clean that readme. This was just for you to reproduce the error and see what the PR fixed

marchdf: ah got it, yeah lets just keep the cpickler modification

marchdf (approved):

PR #273, opened 2022-05-12
Fix Chung corrections

marchdf:
Do not modify viscosity or conductivity if the Chung corrections lead to a negative viscosity or conductivity. Also moved some code into the if condition in case one is computed but not the other.


baperry2 (approved):
Looks good. This will help improve robustness and shouldn't introduce new issues because it only applies the change where the empirical correlation is very clearly wrong.

PR #274, opened 2022-05-12
Add JBBhack

marchdf:
I think it's time to merge this. It lets me run the sCO2 combustor with better mass fraction behavior. Can we find a better name?


esclapez:
Sounds good. As for name:  `ode.OOB_guard` ?

esclapez:
Or rather, `ode.init_oobGuard` since we're only doing that at the start of the integration.

marchdf:
`ode.clean_init_massfrac` ?

esclapez (approved):
It looks good, thanks Marc. I'm fine with it being off by default.

PR #265, opened 2022-04-22
Introducing CEPTR for cantera to Pele C++ mechanisms

marchdf:
I got tired of supporting Pythia and Fuego not being able to parse mechanisms.

Here is a new mechanism converter. Notable things:
- converts Cantera yaml files to mechanism.{cpp,H}
- it is easy to get a cantera yaml file: `ck2yaml --input=mechanism.inp --thermo=therm.dat --transport=tran.dat --permissive` and I will be converting all our files to yaml in the near future
- you just `poetry run convert ${PATH_TO_CANTERA_YAML_FILE}`. I added a Lidryer and air example.
- no more pythia. I want a commit where both exist, so I will be removing that in a separate PR
- By using Cantera, we are using updated values of atomic weights. Pythia uses hard-coded numbers from 1969. So the variations we see in atomic weight will propagate across to other things (e.g. COFETA, etc). This will cause diffs across the codes. To revert back to the old atomic weights, one can add this to the yaml file:
```
elements:
- symbol: C
  atomic-weight: 12.01115
- symbol: O
  atomic-weight: 15.9994
- symbol: H
  atomic-weight: 1.00797
- symbol: N
  atomic-weight: 14.0067
- symbol: AR
  atomic-weight: 39.948
```
- For some mechanisms, cantera has fixed an ambiguity in the way three-body reactions are defined. This will lead to some reordering differences for some mechanisms (e.g. DRM19). More details at https://github.com/Cantera/cantera/issues/1251. This also causes some subtleties when computing the coefficients.
- In doing this, figured out that FORD was not dealt with properly in Fuego and was giving the wrong pre_exponential_factor value for the rates. This can be evident in looking at JL4 mechanism (reaction 0) and noticing that the current value is `39.1` when it should be `219875.458149426`. Verified with Cantera. This happens for mechanisms that have `reaction.orders` as a non-empty dict.
- Also,  reactions with `REV` in Chemkin files is transformed by `ck2yaml` into 2 separate reactions that are irreversible, e.g. NUIGalway:
```
H2O2+H<=>H2O+OH 2.410E+13 0.00 3.970E+03
REV/ 1.269E+08 1.31 7.141E+04 /
```
becomes
```
- equation: H2O2 + H => H2O + OH  # Reaction 31
  rate-constant: {A: 2.41e+13, b: 0.0, Ea: 3970.0}
- equation: H2O + OH => H2O2 + H  # Reaction 32
[...]

PR #275, opened 2022-05-18
Revert atomic weights

marchdf:
Only merge after #265 has shown no/little diffs in our test suites.


malihass:
Importing comments from #265 :

Typo on line 1 of `Testing/Exec/CeptrTest/CanteraSimulation/homogenous-reactor.py`

In the yaml files, if this is not too difficult, can we remove references to mixture state? Otherwise it is ok

In the CEPTR Test, the `utils` folder seems to contain unused files that I was using for other debugging. You can just keep the `initFunction.H` and the `reactFunctions.H`. These are actually already in `IgnitionDelay` test. If we can, removing redundancy would be preferable.

jrood-nrel (approved):

PR #280, opened 2022-05-19
Remove pythia

marchdf:


baperry2:
Do we want to remove the chemkin files (inp/trans/thermo)? These are the originals and the widely accepted standard format. Information shouldn’t be lost in the conversion to yaml, but I don’t know of a tool that converts from yaml back to chemkin if that is desired. For example, if a user wants to run 0D/1D calculations with the same mechanism in software other than cantera, having the chemkin format available would be handy (rather than having to dig into the git history). 

On the other hand, keeping these files might just be clutter that doesn’t get used very often.

marchdf:
This is an interesting point. I removed them to reduce clutter and avoid confusion. My thinking was the yaml encodes in the header where the chem.inp came from and we could track it down that way. But I totally get wanting to keep them. 

I would like a github poll option to know what people would like. I don't care either way really. Let's say upvote to Bruce's comment if you want to keep the chemkin files, downvote his comment to remove them. You have 24h.

marchdf:
🤔  maybe PI counts for 5 votes... But actual developers count for 10.

drummerdoc:
> 🤔 maybe PI counts for 5 votes... But actual developers count for 10.

But GM counts for 20.

baperry2:
> > 🤔 maybe PI counts for 5 votes... But actual developers count for 10.
> 
> But GM counts for 20.

Are these additive? i.e. someone who is GM, PI, and a developer counts as 35 votes?

jrood-nrel:
I vote for starting by keeping them.

jrood-nrel (approved):

marchdf:
This should be good to go. I kept the *.inp and *.dat files.

Issue #282, opened 2022-05-25
Analytical Jacobian fail since CEPTR


malihass:
The solution given by Analytical Jacobian integration seems different than for finite difference Jacobian since the move to CEPTR

Attached is the Solution given for the same 0D case with `dodecane_lu` 

![dodecane_beforeCEPTR](https://user-images.githubusercontent.com/5544573/170306427-0a71c422-c1f1-4de2-90c8-fcece41da607.png)

![dodecane_afterCEPTR](https://user-images.githubusercontent.com/5544573/170306369-4e28ba61-edbd-4399-ba19-2c7ab03458f8.png)

Same problem seem to occur with `drm19` but not `LiDryer`. Currently we think the third-body reactions could be the issue

malihass:
#283 fixed `drm19` and `Lidryer`. Still debugging some lingering issues with `dodecane_lu`

malihass:
We found this was a compiler related issue. Clang 11 finds a false positive fpe in the analytical jacobian computation. After deactivating the `amrex_fpe_trap`, we found that the test passes. The problem does not occur with clang13. 

We verified that the switch to CEPTR does not change the values of `wdot`, Jacobian, and preconditioning Jacobian, with `consP=1` and `consP=0`.

Here are the new test results with `dodecane_lu` after the switch to CEPTR.    
![Figure_1](https://user-images.githubusercontent.com/5544573/171685243-ee4cbaf3-d4e9-4cb8-8060-9d288f49ea89.png)

drummerdoc:
That looks great!  It is possible to pick a spot where it's tricky (like 0.1s, say) and do a single-step solve at a pretty difficult dt and print out enough stuff to see if the Newton solve is converging quadratically?

PR #283, opened 2022-05-25
Fix ajac generation from CEPTR

marchdf:
Related to #282


malihass (approved):

malihass:
This fixed `drm19` and `Lidryer` issues, analytical Jacobian and wdot are the same up to machine precision.
There is a lingering issues with dodecane_lu. Thanks Marc!

PR #285, opened 2022-06-01
correct species init to stay bounded in Eos and Tran Eval utilities

baperry2:
Previous init causes out of bounds species.


marchdf:
Awesome! Thank you! 2 comments:
- can you do the clang format thing? 
- can you see if the same thing needs to happen to the ReactEval case?

baperry2:
Fixed the formatting.

This should cause failures for the Eos and Transport tests.

marchdf (approved):

PR #290, opened 2022-06-02
fix computation of sc_qss for aJacobian

malihass:
The analytical Jacobian was full of nans when using a QSS mechanism because the `sc_qss` were not computed correctly. The function `comp_qss_coeff` needs to be called before `comp_sc_qss`


malihass:
This is a 0D ignition case I ran for `dodecane_lu_qss` 
The Analytical Jacobian method can be very sensitive to `dt`. If we add enough steps, we do match the finite difference Jacobian. This suggests that we will need to correctly chain rule the `sc_qss` instead of assuming that they are constant at every step. At least this shows consistency with finite difference Jacobian for small enough dt.
Figure_1

marchdf (approved):

PR #295, opened 2022-06-06
More code fixes

jrood-nrel:


Eos/GammaLaw.H
```diff
@@ -113,7 +113,7 @@ struct GammaLaw
   static void RYET2P(
     amrex::Real R,
     amrex::Real* /*Y*/,
-    amrex::Real& E,
+    const amrex::Real& E,
```
marchdf: I don't think we should pass `E` by reference. But I like the `const`. I think this comment goes for several of the edits below.

Reactions/ReactorCvode.cpp
```diff
@@ -567,8 +567,8 @@ ReactorCvode::checkCvodeOptions() const
     PS = SUNSparseMatrix(
       (NUM_SPECIES + 1), (NUM_SPECIES + 1), nJdata, CSR_MAT,
       *amrex::sundials::The_Sundials_Context());
-    int* rowCount = (int*)SUNSparseMatrix_IndexPointers(PS);
-    int* colIdx = (int*)SUNSparseMatrix_IndexValues(PS);
+    int* rowCount = static_cast<int*>(SUNSparseMatrix_IndexPointers(PS));
```
marchdf: in another commit somewhere, you also did `auto*` to avoid another warning type. Do this here and in the line below?

Transport/Simple.H
```diff
@@ -28,7 +28,7 @@ struct NonIdealChungCorrections<eos::SRK>
     const amrex::Real Temp,
     const amrex::Real* Xloc,
     const amrex::Real rholoc,
-    amrex::Real& wbar,
+    const amrex::Real& wbar,
```
marchdf: Same here, remove the reference on `wbar`?

marchdf (commented):
A couple of minor things maybe?

marchdf (approved):

PR #298, opened 2022-06-09
Use std move in plot file manager

jrood-nrel:


marchdf (approved):

Utility/PltFileManager/PltFileManager.cpp
```diff
@@ -21,8 +22,8 @@ GotoNextLine(std::istream& is)
   is.ignore(bl_ignore_max, '\n');
 }
 
-PltFileManager::PltFileManager(const std::string& a_pltFile)
-  : m_pltFile{a_pltFile}
+PltFileManager::PltFileManager(std::string a_pltFile)
```
marchdf, line 25: can we still keep this `const`?

jrood-nrel: Haha, no it can't

PR #300, opened 2022-06-23
ensure RK64 reactor doesn't integrate beyond the desired time

baperry2:
The RK64 integrator just goes until the desired time is reached or exceeded and the timesteps are generally not even fractions of the integration time, so the last timestep will often go past the desired time by a small amount.

This PR fixes that by ensuring that the last timestep gets you exactly to the desired integration time.


hsitaram (approved):
Looks good! very nice trick

PR #305, opened 2022-07-14
Charged species in CEPTR

esclapez:
Add a couple of things to handle changed species in ceptr:
 - speciesDb now knows about species charges
 - count the number of ions to provide a NUM_IONS macro
 - add function to species unit charge and charge per unit mass


Support/ceptr/ceptr/converter.py
```diff
@@ -456,11 +459,17 @@ def mechanism_header_includes(self, fstream):
...
             )
+            if s[-1] == "n" or s[-1] == "p" or s == "E":
+                nb_ions += 1
         cw.writer(fstream)
         cw.writer(fstream, "#define NUM_ELEMENTS %d" % (nb_elem))
         cw.writer(
             fstream, "#define NUM_SPECIES %d" % (self.species_info.n_species)
         )
+        if nb_ions > 0:
+            cw.writer(
```
marchdf: Do we just always want to define this for all mechs? If the code is using `NUM_IONS` anywhere it won't compile with any of the other mechs. So maybe all mechs should have this defined?

marchdf (approved):

Support/ceptr/ceptr/converter.py
esclapez: You're probably right. I didn't want to add it to all the mechs that do no have charged species, but it's probably safer to do it this way.

marchdf: yeah it's not great but it feels safer... maybe there's some other way? I can't think of one...

marchdf (approved):

PR #309, opened 2022-07-23
Symbolic computation of Jacobian of QSS mechanisms

malihass:
1)Use `sympy` and `symengine` to symbolically encode the production rates.
2)Allow the user to print symbolic Jacobian in place of the old Jacobian. Effect is shown below. 
Left was old Jacobian implementations (works for very small timesteps). Right is the new one.

aj_0D_QSS

This should also help address the GMRES instabilities that we noticed. Below is what happens with finite difference Jac, analytical Jac and GMRES for the same 0D test case with same tolerances and timestepping

qss_integrator

3)Print the Jacobian using common sub-expression eliminations
4)Do some management of common subexpression to reduce the memory footprint. Format options can be controlled by the user via `.toml` file
5) Update documentation to explain how this procedure works.
6) The `mechanism.H` for `dodecane_lu_qss` is the one with symbolically computed Jacobian.

Co-authored-by: Nicholas Wimer [nicholas.wimer@gmail.com](mailto:nicholas.wimer@gmail.com)


Support/ceptr/makeQss_toml.sh
```diff
@@ -0,0 +1,3 @@
+poetry run qssa -f ${PELE_PHYSICS_HOME}/Support/Mechanism/Models/dodecane_lu_qss/skeletal.yaml -n ${PELE_PHYSICS_HOME}/Support/Mechanism/Models/dodecane_lu_qss/non_qssa_list.yaml
```
marchdf: remove this file and just put the example commands in the documentation.

Support/ceptr/qssa_input.toml
```diff
@@ -0,0 +1,23 @@
+[Readability]
```
marchdf: Remove this file and just put it in the documentation.

Support/ceptr/ceptr/species_info.py
```diff
@@ -61,3 +63,254 @@ def __init__(self):
...
+        numb_list = []
+        symb_list = []
+
+        for idx in range(self.n_species):
+            name_list.append(f"sc[{idx}]")
+            numb_list.append(idx)
+            symb_list.append(self.nonqssa_species_list[idx])
+
+        # Initialize the dataframe
+        sc_df = pd.DataFrame(
```
marchdf, line 267: Can this be a plain `OrderedDict`? That way we don't have to import/install/depend on `pandas`. Just a thought...

Support/ceptr/ceptr/ceptr.py
```diff
@@ -1,45 +1,91 @@
...
+        type=str,
+        default=None,
+        required=False,
+    )
+
+    parser.add_argument(
+        "-sj",
+        "--symbolic_jacobian",
+        action="store_true",
+        help="Compute the Jacobian using symbolic recording",
```
marchdf: Is this only for QSS? Maybe add that to the help string? Maybe the option should be `qss_symbolic_jacobian`?

Support/ceptr/ceptr/ceptr.py
```diff
@@ -1,45 +1,91 @@
...
+        description="Mechanism converter", formatter_class=RawTextHelpFormatter
+    )
     group = parser.add_mutually_exclusive_group(required=True)
     group.add_argument("-f", "--fname", help="Mechanism file", type=str)
     group.add_argument(
         "-l", "--lst", help="Mechanism directory file list", type=str
     )
+
+    parser.add_argument(
+        "--format_input",
```
marchdf: can we call this option `qss_jacobian_format` ?

Support/ceptr/ceptr/converter.py
```diff
@@ -2,17 +2,20 @@
 import pathlib
 import shutil
 import subprocess as spr
+import time
```
marchdf: Remove all `time` stuff. We can just use a profiler.

Support/ceptr/ceptr/species_info.py
marchdf: Chatted with Nick and I am convinced that pandas should stay.

esclapez:
@malihass I'll give it a try on a case today. Can't say much about the python :)

marchdf (approved):

esclapez (approved):
Successfully tested the dodecane_qss mechanism on an single igniting dodecane jet, and the Jacobian seems to work nicely !

PR #316, opened 2022-08-10
Better global launches

marchdf:


marchdf:
So Paul found a 10% speedup using `nbThreads = 64` on Crusher with this code on the challenge problem. The part I am not sure about is this:
```
__attribute__((noinline))
```
on RTY2JAC. It helps register spillage on AMD we think. Should we make that depend on the compiler? It's not clear what nvcc is doing with this code.

We need to test this a bit more...

esclapez:
Thanks Marc ! I'll give that a try before approving to get a figure on the LM(eX) side.

marchdf:
yeah that would be great! It's nothing crazy on the C side...

Reactions/ReactorCvodeJacobian.cpp
```diff
@@ -40,16 +40,54 @@ cJac(
...
-                 stride = blockDim.x * gridDim.x;
-             icell < ncells; icell += stride) {
-          fKernelComputeAJchem(
-            icell, NNZ, react_type, csr_row_count_d, csr_col_index_d, yvec_d,
-            Jdata);
-        }
-      });
+
+    if (nbThreads <= 32) {
+      amrex::launch_global<32><<<nbBlocks, nbThreads, ec.sharedMem, stream>>>(
```
jrood-nrel: Should we get rid of the `ifs` here and use `CVODE_NB_THREADS`?

marchdf: We could do that. And then remove the user parameter and just make it a compile time thing?

jrood-nrel: I'm thinking so.

marchdf: Ok I did that.

jrood-nrel (approved):

esclapez:
So tested with the 52 species dodecane (to stress the Jacobian more) and the single jet ignition case:
 - before this PR:
 `Pele::ReactorCvode::cJac()                     12712      105.8      129.1      155.1  16.35%`
 - after this PR:
 `Pele::ReactorCvode::cJac()                     12712      42.49     51.87      62.69   7.29%`
So a definite win on the LM(eX) side with bigger time steps. I'll approve and do some more tests on the challenge problem with the QSS mech.
Side question: do we thing the RHS kernel could also benefit from this ?

esclapez (approved):

marchdf:
Awesome! Thanks for testing! This is more dramatic than we are seeing in PeleC. So in that sense I am not sure if we could measure anything meaningful in PeleC for the RHS idea. The RHS already uses a parallelfor (instead of the launch global) so that takes care of the number of threads business. Though we could reduce the amrex global number of threads and see what happens. That being said, it might be worth adding `__attribute__((noinline))` to `RTY2WDOT` for AMD and trying it with LM. The register pressure there is much less so we might not see anything... 

 I just found another launch_global in the preconditionner that could benefit from this so I added that.

marchdf:
Are we good to merge this then? I am happy with this as is.

PR #328, opened 2022-10-20
Change conditional statement to handle special efficiencies case (fix #327)

### LINKED ISSUE #327: Special case for single non-unity third-body efficiency
Hi,

I stumbled upon some weird behaviour when using CEPTR to convert Cantera mechanism files to the format used in PeleLM. 

So, for example  in `Support/ceptr/utilities.py`, if the reaction has only one efficiency the third body is added to the reactants and products : 
```python
if hasattr(reaction, "efficiencies"):
        if len(reaction.efficiencies) == 1:
            reagents = copy.deepcopy(
                dict(
                    sum(
                        (
                            Counter(x)
                            for x in [reagents, reaction.efficiencies]
                        ),
                        Counter(),
                    )
                )
            )
```

This works well for reactions like reaction 35 in **grimech12** (H + O2 + H2O <=> HO2 + H2O) because H2O is added to the reactants and products.

But I am converting a mechanism where there is a third body reaction such as : 

NO2 + M <=> NO + O + M  (see Reaction 52 in [Jiang.yaml](https://github.com/AMReX-Combustion/PelePhysics/files/9816498/Jiang.txt))

With only H2O specified in the efficiences : 

`efficiencies: {H2O: 16.25}`

In this cases the reaction is re-defined as : 

NO2 + 16.25 H2O <=> NO + O + 16.25 H2O

Which alters the results when comparing PeleLM and Cantera data.

I feel like this is a special case where things break  because modifying CEPTR to handle the special case  : 

`len(reaction.efficienties) == 1 AND reaction.efficienties.values != 1.0 `

produces the same mechanism.H and mechanism.cpp files for all the mechanism shipped with PeleLM. 

I have put the special case handling in a fork : [olivecha/PelePhysics/ceptr-fix](https://github.com/olivecha/PelePhysics/tree/ceptr-fix)

with a [regression test](https://github.com/olivecha/PelePhysics/blob/ceptr-fix/Support/ceptr/tests/test_regression.py) to ensure changes do not affect the generation of the supplied mechanisms.

olivecha:


marchdf:
If approved it would:

Closes #327

Support/ceptr/tests/test_regression.py
```diff
@@ -0,0 +1,81 @@
+import os
```
marchdf: Can you remove this file from the PR? It's a great idea but I think best done as a separate thing.

olivecha: I removed it from the ceptr-fix branch.

malihass:
I can see that the fix does help match Cantera, however the difference is much more subtle in 0D reactor than what you showed with the 1D flame

Before

withoutBugFix

After 

withBugFix

Also looking at the `mechanism.H` spit out by CEPTR for the mechanism you linked, I can see that CEPTR considers that efficiencies of all the species is 0 except that of the species indicated (H2O here). This is definitely wrong so thanks for looking into this :-)

I am however puzzled by the fix. To me, the issue is that in the test that checks `if number of efficiency==1:` we need to add `and that the efficiency species is not part of the reactants` (that is that efficiency is that of a species in a third body reactant). In other terms, the issue has nothing to do with the value of the efficiency factor.

I propose to change all the `len(reaction.efficiencies) == 1` into `len(reaction.efficiencies) == 1 and (list(reaction.efficiencies.keys())[0] in list(reaction.reactants.keys()))`. I tested it and it gives the same results as the current fix of @olivecha for this mechanism. 

@olivecha and @marchdf do you agree, or am I missing something?

marchdf:
Thank you for looking into this @malihass ! I like your proposed fix. @olivecha do you want to try that and see if that works and then update the PR? 

On another note, we do some linting and formatting checks on python. I recommend you `poetry run black .` and same with isort and flake8 to automatically fix the code.

malihass:
@olivecha  : see the branch https://github.com/AMReX-Combustion/PelePhysics/tree/thirdBody_eff for the candidate alternative fix

olivecha:
Hi, 

The alternative fix seems to change the output for the mechanisms already present in PelePhysics. For example, the line 3867 and 3868 in mechanism.H for grimech12 changes from : 

```cpp
  // reaction 34: H + O2 + H2O <=> HO2 + H2O
  qf[25] = sc[1] * sc[3] * sc[5];
  qr[25] = sc[5] * sc[6];
```
to : 

```cpp
  // reaction 34: H + O2 + H2O <=> HO2 + H2O
  qf[25] = sc[1] * sc[3];
  qr[25] = sc[6];
```
I think this is because Cantera removes the species present in both the reactants and the products from the equation so the condition is never true. 

Instead of checking that the efficiency is equal to one to determine if the reaction has a specific third body, I think the reaction attribute ```reaction.default_efficiency``` could be used. It is equal to zero for reaction with a defined third body and one if only one efficiency is supplied. The condition would then be : 

```len(reaction.efficiencies) == 1 and isclose(reaction.default_efficiency, 0.0)```

malihass:
I agree with your latest changes, `reaction.default_efficiency` is the right way of checking whether the efficiency is due to a third body. I checked that the 0D reactor test passed. I also checked that this did not change any of the other mechs. Only the `NUIGalway` mechanism is affected by this change. The efficiency of the TROE falloff were not correctly treated before and are now correct. 

Great job @olivecha !

malihass (approved):

PR #330, opened 2022-11-04
Soret effect implementation

ThomasHowarth:
Additions to PelePhysics for obtaining thermal diffusion coefficients for Soret effects - see corresponding pull request for PeleLMeX. @nickwimer @drummerdoc @esclapez @marchdf


Transport/Simple.H
```diff
@@ -391,6 +413,7 @@ struct SimpleTransport
     const bool wtr_get_mu = true;
     const bool wtr_get_lam = true;
     const bool wtr_get_Ddiag = true;
+    const bool wtr_get_chi = true;
```
marchdf: Should this default to false?

Transport/Simple.H
```diff
@@ -379,6 +400,7 @@ struct SimpleTransport
     amrex::Array4<const amrex::Real> const& T_in,
     amrex::Array4<const amrex::Real> const& Rho_in,
     amrex::Array4<amrex::Real> const& D_out,
+    amrex::Array4<amrex::Real> const& chi_out,
```
marchdf, line 397: So this changes the call signature of `get_transport_coeffs`. I think it would be necessary to update the other transport types to have the same signature that way the call site doesn't take an if condition.

Transport/Simple.H
```diff
@@ -405,15 +428,16 @@ struct SimpleTransport
 
           amrex::Real muloc, xiloc, lamloc;
           amrex::Real Ddiag[NUM_SPECIES] = {0.0};
-          transport(
-            wtr_get_xi, wtr_get_mu, wtr_get_lam, wtr_get_Ddiag, T, rho, massloc,
-            Ddiag, muloc, xiloc, lamloc, tparm);
+          amrex::Real chi_mix[NUM_SPECIES] = {0.0};
```
marchdf, line 425: we might want to wrap this definition and chi_out with some if conditions based on whether they are asked for. I guess the issue is that our current call doesn't let the user do that. By default it asks for all coeffs. Maybe that needs to change. Thoughts anyone?

Transport/Simple.H
```diff
@@ -370,6 +372,25 @@ struct SimpleTransport
     if (wtr_get_Ddiag) {
       BinaryDiff<EosType>()(Xloc, Yloc, logT, rholoc, Tloc, Ddiag, tparm);
     }
+    
+    if (wtr_get_chi) {
+      int four_idx_nj;
```
marchdf: move definition of this into line 383 and const it. Same for `chi`.

marchdf (changes requested):

Transport/Simple.H
drummerdoc: Agreed, we can just not write to that variable if it is not asked for.  Then the use can pass a dummy scalar in that slot and all will be well.

Transport/Simple.H
ThomasHowarth: Added the appropriate call signatures to constant and sutherland transport types,  assume they're the required changes?

Transport/Simple.H
ThomasHowarth: The only way I can see to not get chi would be to grab the parmparse flag in the transport params init and check the tparm here? is that a reasonable solution? currently it is caught under a pelelm.use_soret flag but I could change to transport.use_soret (with some appropriate changes in the pelelmex PR)

Transport/Simple.H
marchdf: yes, perfect.

Transport/Simple.H
marchdf: I kinda like this idea. Happy to hear alternatives from others as well. But I would vote for your suggestion.

drummerdoc: Either would work, and the current version is fine.  It kinda makes sense to have this under transport-specific flags, but that would require that C, LM and LMeX make a bunch of changes that will break everyone's input files, and that's probably a bad idea.

marchdf: If it's an optional transport flag that defaults to false, then I think it won't require any changes. At least on the C end.

ThomasHowarth: I've added in the call in the transport params init to check if use_soret is flagged and adjusted the wtr_get_chi boolean to check this instead (also addresses the comment above) - let me know if this looks good

Transport/TransportParams.H
```diff
@@ -87,6 +87,10 @@ struct TransParm<EOSType, SimpleTransport>
   amrex::GpuArray<amrex::Real, NUM_SPECIES* NUM_FIT> fitlam = {0.0};
   amrex::GpuArray<amrex::Real, NUM_SPECIES* NUM_SPECIES* NUM_FIT> fitdbin = {
     0.0};
+  int use_soret = 0;
```
marchdf: can we make this a bool?

Transport/Constant.H
```diff
@@ -66,20 +74,23 @@ struct ConstTransport
     const bool wtr_get_mu = true;
     const bool wtr_get_lam = true;
     const bool wtr_get_Ddiag = true;
+    const bool wtr_get_chi = false;
```
marchdf: can this (and sutherland) also use the tparm to get the value? Just to avoid any confusion down the line.

Transport/Simple.H
```diff
@@ -370,6 +372,23 @@ struct SimpleTransport
     if (wtr_get_Ddiag) {
       BinaryDiff<EosType>()(Xloc, Yloc, logT, rholoc, Tloc, Ddiag, tparm);
     }
+    
+    if (wtr_get_chi) {
+      for (int n = 0; n < tparm->numLite; ++n) {
+	int i = tparm->liteSpec[n];
```
marchdf: const this `int`

marchdf (changes requested):

Transport/Constant.H
ThomasHowarth: For constant transport I've add the option of constant thermal diffusion ratios for consistency, although the physical implications are a little odd. For sutherland I've just hardcoded to false and not asked for it.

marchdf:
@ThomasHowarth thanks for all the work! Now we just need to get the CI to pass. The first thing is to run `clang-format` on the code you changed. If you go to the `Transport` source directory, you can run: `find . \( -name "*.cpp" -o -name "*.H" \) -exec clang-format -i {} +` (you need a recent version of clang-format to do this.

marchdf:
@ThomasHowarth The next step is to update the `get_transport_coeffs` calls that are in the Transport tests ;)

Transport/TransportParams.H
```diff
@@ -105,7 +117,11 @@ struct InitTransParm<EOSType, SimpleTransport>
     egtransetCOFETA(tparm->fitmu.data());
     egtransetCOFLAM(tparm->fitlam.data());
     egtransetCOFD(tparm->fitdbin.data());
-
+    amrex::ParmParse pp("transport");
+    pp.query("use_soret", tparm->use_soret);
+    egtransetNLITE(&tparm->numLite);
```
baperry2: So if I'm understanding things correctly you allocate space for the parameters needed for Soret for up to 10 light species, and then using this function the mechanism defines how many light species it actually has. 

If that's accurate, you should add something that throws an error if the mechanism indicates more light species than space was allocated for. Also can the default be reduced to 3 because it looks like that's the most in any of our mechanisms?

Transport/Simple.H
```diff
@@ -405,15 +427,16 @@ struct SimpleTransport
...
           transport(
-            wtr_get_xi, wtr_get_mu, wtr_get_lam, wtr_get_Ddiag, T, rho, massloc,
-            Ddiag, muloc, xiloc, lamloc, tparm);
+            wtr_get_xi, wtr_get_mu, wtr_get_lam, wtr_get_Ddiag, wtr_get_chi, T,
+            rho, massloc, Ddiag, chi_mix, muloc, xiloc, lamloc, tparm);
 
           //   mu, xi and lambda are stored after D in the diffusion multifab
           for (int n = 0; n < NUM_SPECIES; ++n) {
             D_out(i, j, k, n) = Ddiag[n];
+            chi_out(i, j, k, n) = chi_mix[n];
```
baperry2, line 433: strictly speaking this could also be deactivated if use_soret is false, but probably not a big deal either way

baperry2 (commented):
Thanks for the PR to add this capability!

One thing I'd like to see is a bit of documentation/commenting describing the new capability and implementation (e.g., what physical assumptions, what flags are needed to activate, what happens for each type of transport model, etc). 

I know the current state of the documentation is pretty incomplete, but if you can figure out a place to slide that into the Readthedocs or Readme, I'm sure it would be a big help to anyone seeking to use this capability in the future.

Transport/TransportParams.H
ThomasHowarth: I'll put a flag in to catch this potential memory issue - as for the default number the only common species that will ever be classed as light will be H, H2 and He. Only exceptions I can think of are isotopes of hydrogen, and I'm also not sure how the mechanism deals with the case of electron transport whens compiled with electric field capabilities. Either way, I can almost certainly reduce this number, and it would be nice if the NLITE was #define'd in mechanism, so this could be hardcoded at compile time rather than the way it is currently done.

ThomasHowarth:
> Thanks for the PR to add this capability!
> 
> One thing I'd like to see is a bit of documentation/commenting describing the new capability and implementation (e.g., what physical assumptions, what flags are needed to activate, what happens for each type of transport model, etc).
> 
> I know the current state of the documentation is pretty incomplete, but if you can figure out a place to slide that into the Readthedocs or Readme, I'm sure it would be a big help to anyone seeking to use this capability in the future.

I'd be happy to - would this go into PelePhysics or PeleLMeX documentation? The PelePhysics documentation looks to be mostly chemistry focussed, and I can't actually find anywhere that talks about the different transport models other than setting them as a compile option.

Transport/TransportParams.H
baperry2: So it's pretty simple to modify CEPTR to add a define for NLITE (see here: https://github.com/AMReX-Combustion/PelePhysics/commit/6bb26e3d5a151933ca2953a7de84bca177b4bf28). If you go this route, the mechanism source files will need to be regenerated using CEPTR (@marchdf would probably be able to help out with that if needed) in order to use this functionality. But also just reducing the preset to 3 and adding some error capture is fine.

baperry2:
> I'd be happy to - would this go into PelePhysics or PeleLMeX documentation? The PelePhysics documentation looks to be mostly chemistry focussed, and I can't actually find anywhere that talks about the different transport models other than setting them as a compile option.

Good point. Maybe just throw a line in the PelePhysics Readme where it talks about transport models to say that it can optionally calculate coefficients for Soret, and then put a bit more detail in the LMeX docs when you finish up that PR.

ThomasHowarth:
Sorry for the delay on the last couple changes. I've reduced the number of light species down to 3 - if we think it's worth hardcoding the number of light species in the mechanism down the line we can do so. I've also added a line to the README saying that mixture-averaged thermal diffusion coefficients are available using the appropriate flags, and referred to the undergoing LMeX PR. If you'd like more information/different wording, let me know. Is there anything else required from me? I see the tests fail on some mechanisms, do they just need regenerating?

esclapez (approved):
Sorry it took so long on my end. I'll make sure all the tests are passing and merge it. Thanks for doing this, it's great ! I'll update the PeleLMeX side when this is merged.

marchdf (approved):

baperry2 (approved):

PR #331, opened 2022-11-18
Doc update

marchdf:
Moved some stuff out of PeleC to document it here. And then I reorganized some of the sections. Nothing major but in going through this there is a lot that needs to be freshened up... including discussions about cvode/sundials... Anyway this is just a small pass at things.


baperry2 (approved):
Good start on reorganizing this documentation. EOS stuff definitely makes more sense here than in the PeleC docs. There's definitely far more work to do on this beyond the scope of this PR.

PR #333, opened 2022-12-08
Solve warnings on new Github virtual environments

jrood-nrel:
This also fixes out of bounds issues with 2D turbinflow, though I'm guessing 2D turbinflow isn't intended.


esclapez (approved):

esclapez:
It is indeed not intended for 2D in the current state, but I actually added it to my TODO list. It could be useful.

PR #334, opened 2022-12-08
Use Cray libsci library for MAGMA

jrood-nrel:
I'm thinking we're going to want to use `PrgEnv-cray` on Frontier to have the most support and this is necessary for that to build. I'm going to make it a draft at the moment though.


esclapez (commented):
I have no way of testing this at this point, but go ahead.
Can you also update the magma version to 2.7.0, I was testing on Perlmutter and it seemed to alleviate some of the OOM issues I was seeing before.

jrood-nrel:
This ran on Crusher with the challenge problem so I'm fine updating MAGMA.

esclapez (approved):

PR #346, opened 2023-01-04
add RTY2H for SRK (and other EOS)

baperry2:
We previously did not support RTY2H EOS calls, only TY2H which is only applicable for ideal EOSs. This PR add an RTY2H function for SRK, corrects the SRK TY2H to throw an error, and adds RTY2H wrappers around TY2H for the other EOSs.

It turns out that RTY2H is easy to implement through the definition h = e + p/rho because we already have functions to compute e and p (@shashankNREL please verify I'm not missing anything).


shashankNREL (approved):
Looks good to me. Thanks

PR #350, opened 2023-02-06
Hari/bdfscheme

hsitaram:
Adding backward-difference scheme for chemistry integration, everything hand-rolled including a preconditioned GMRES.
Hope this will be useful for probing into finding the right pre-conditioners and extension to new architectures with basic AMReX code. 

some notes and thoughts that will need some future efforts:

1) usage from any application code: do chem_integrator="ReactorBDF"

2) Unlike Cvode, there is no internal estimation of time-step. I did some tests ~ 10-50 substeps per microsec seems to work for the most-often used mechanisms (like Lidryer, drm19, dodecane_lu), but this will be something more to do.

3) ReactorBDF seems ~ 2X faster on CPU than Cvode for the reacteval case giving the same solution, but this number will change based on all the cvode parameters (tols etc..). It is slower than Cvode on GPU though, I am looking at this now.

### DIFF
--- a/Testing/Exec/IgnitionDelay/inputs/inputs.0d_firstpass
+++ b/Testing/Exec/IgnitionDelay/inputs/inputs.0d_firstpass
@@ -32,6 +32,8 @@ amrex.fpe_trap_invalid = 1
 amrex.fpe_trap_zero = 1
 amrex.fpe_trap_overflow = 1
 
+ode.bdf_nsubsteps=2000
+ode.bdf_scheme=3
 
 # 1013250.0 * patm
 hr.press = 5066250.0
--- a/Testing/Exec/ReactEval/inputs.3d_1dArray
+++ b/Testing/Exec/ReactEval/inputs.3d_1dArray
@@ -5,8 +5,8 @@ fuel_name = CH4
 reactFormat = 1dArray
 chem_integrator = "ReactorRK64"
 
-ode.dt  = 1.e-05
-ode.ndt = 1
+ode.dt  = 2.e-05
+ode.ndt = 30
 ode.ode_ncells = 1024
 ode.reactor_type  = 1 # 1=full e, 2=full h
 ode.verbose = 1
@@ -37,3 +37,5 @@ cvode.max_order = 4
 amrex.fpe_trap_invalid = 1
 amrex.fpe_trap_zero = 1
 amrex.fpe_trap_overflow = 1
+
+ode.bdf_nsubsteps=40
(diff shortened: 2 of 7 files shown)


Testing/Exec/ReactEval/inputs.3d_1dArray
```diff
@@ -5,8 +5,8 @@ fuel_name = CH4
 reactFormat = 1dArray
 chem_integrator = "ReactorRK64"
 
-ode.dt  = 1.e-05
-ode.ndt = 1
+ode.dt  = 2.e-05
```
marchdf, line 8: Are the changes in this file and in the other ReacEval input file deliberate? Do we want a ReactorBDF input file too?

marchdf (approved):

marchdf:
Main comment: does it work on GPU?

Testing/Exec/ReactEval/inputs.3d_1dArray
hsitaram: yes, without these, even cvode was failing to integrate this case

hsitaram:
Yes, tested on GPU also. No problems

malihass:
I tried it on the ignition delay case
`tIgn = 0.0763 +/- 2.09e-05 s` with CVODE
`tIgn = 0.0765 +/- 2.12e-05 s` with BDF
So I would say not bad!
The execution was a lot slower with BDF (about 2 orders of magnitude on my laptop) with 1e-3s per step.

Testing/Exec/IgnitionDelay/inputs/inputs.0d_firstpass
```diff
@@ -32,6 +32,8 @@ amrex.fpe_trap_invalid = 1
 amrex.fpe_trap_zero = 1
 amrex.fpe_trap_overflow = 1
 
+ode.bdf_nsubsteps=2000
```
malihass, line 35: Is this the same as `ndt` ? If yes, can we call it the same as `ndt` ?

hsitaram: This is not same as ndt. ndt is specific to this test case and is not there in PeleC/LM/LMeX. see https://github.com/AMReX-Combustion/PelePhysics/blob/cc48e5cfa352ae0479d402735b6bf9097efb612e/Testing/Exec/IgnitionDelay/utils/initFunctions.H#L67

nsubsteps is a way to set the timestep internally when BDF does its advance.

malihass (approved):

esclapez:
Successfully tested in LMeX, on the FlameSheet case. I had to increase the nsubstep to get a stable integration with the large low Mach step size, but it still worked nicely !

esclapez (approved):

PR #351, opened 2023-02-07
Check for PELE_COMPILE_AJACOBIAN, Add print profiling option

gardner48:
* Check if `PELE_COMPILE_AJACOBIAN` is True and add directive
* Don't print profiling information for each integration by default

### DIFF
--- a/Reactions/ReactorCvode.H
+++ b/Reactions/ReactorCvode.H
@@ -158,7 +158,7 @@ private:
 #endif
   int max_nls_iters{3};
   int max_fp_accel{2};
-  int print_profiling{0};
+  int m_print_profiling{0};
   bool m_clean_init_massfrac{false};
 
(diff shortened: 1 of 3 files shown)


esclapez (approved):
Looks good. Thanks for adding the define for the Jacobian David !

Reactions/ReactorCvode.H
```diff
@@ -158,6 +158,7 @@ private:
 #endif
   int max_nls_iters{3};
   int max_fp_accel{2};
+  int print_profiling{0};
```
marchdf: would you mind prefixing with `m_` ? I've been trying to do that for member variables (obviously haven't been able to do that)

marchdf (changes requested):

marchdf (approved):

PR #356, opened 2023-02-13
Fix some warnings

marchdf:


hsitaram (approved):
all good! not sure, if any application (C/LM/ReactEval etc..) uses the MOD_REACTOR flag. we should probably remove it if its not used.

marchdf:
`ReactEval` and Co seem to use it... not sure the point of it though?

jrood-nrel (approved):

PR #357, opened 2023-02-15
Jacobian transpose

malihass:
This is the implementation of the row major Jacobian we discussed.

## CEPTR

- pass an option `-rj`  to signify that the Jacobian needs to be transposed
- Change the name of `DWDOT` to `DWDOT_ROLL` so that the transposed Jacobian can be used only if the correct flags are used (`PELE_ROLL_JAC`)
- When we transpose the Jacobian, we transpose also the precond Jacobian

## EOS

- Call `DWDOT_ROLL` if  `PELE_ROLL_JAC` flag is used. This makes sure that `Reactors` that are fed a transposed Jacobian expect it. Otherwise it would throw an error.

## Reactor

- Transpose the indices when using Jacobian

## Validation

I run a 0D reactor with `dodecane_lu` and `dodecane_lu_qss` (symbolic Jacobian + memory optimization) with CVODE and AJ. I run the transposed and non transposed version. The results seem consistent. The Cantera line is integration of the skeletal mechanism with Cantera. 

Screen Shot 2023-02-15 at 5 04 45 PM

### DIFF
--- a/Reactions/ReactorBDF.cpp
+++ b/Reactions/ReactorBDF.cpp
@@ -69,13 +69,13 @@ get_bdf_matrix_and_rhs(
   for (int ii = 0; ii < NUM_SPECIES; ii++) {
     for (int jj = 0; jj < NUM_SPECIES; jj++) {
       Jmat2d[ii][jj] = -bdfp.FCOEFFMAT[tstepscheme][0] *
-                       Jmat1d[jj * (NUM_SPECIES + 1) + ii] * mw[ii] / mw[jj];
+                       Jmat1d[ii * (NUM_SPECIES + 1) + jj] * mw[ii] / mw[jj];
     }
     Jmat2d[ii][NUM_SPECIES] = -bdfp.FCOEFFMAT[tstepscheme][0] *
-                              Jmat1d[NUM_SPECIES * (NUM_SPECIES + 1) + ii] *
+                              Jmat1d[ii * (NUM_SPECIES + 1) + NUM_SPECIES] *
                               mw[ii];
     Jmat2d[NUM_SPECIES][ii] = -bdfp.FCOEFFMAT[tstepscheme][0] *
-                              Jmat1d[ii * (NUM_SPECIES + 1) + NUM_SPECIES] /
+                              Jmat1d[NUM_SPECIES * (NUM_SPECIES + 1) + ii] /
                               mw[ii];
   }
   Jmat2d[NUM_SPECIES][NUM_SPECIES] =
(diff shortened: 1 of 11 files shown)


hsitaram:
Before we merge this

1) can you check if you are using AJ for the ignitiondelay case. I think you need to set the solver to be dense_AJdirect instead of dense_direct

2) @gardner48 or @balos1 - can you tell us if the jacobian is stored as row or column major in sundials?

Reactions/ReactorBDF.cpp
```diff
@@ -65,6 +65,28 @@ get_bdf_matrix_and_rhs(
   get_rho_and_massfracs(soln, rho, massfrac);
 
   eos.RTY2JAC(rho, soln[NUM_SPECIES], massfrac, Jmat1d, consP);
+#ifdef PELE_ROLL_JAC
```
marchdf, line 68: @jrood-nrel should we just use if conditions? feels more natural and removes a compile option.

marchdf:
Very cool! Can we try a performance test and see if there is any benefit?

malihass:
> 1. can you check if you are using AJ for the ignitiondelay case. I think you need to set the solver to be dense_AJdirect instead of dense_direct

Yes I am using `denseAJ_direct` for the validation cases I showed

Reactions/ReactorBDF.cpp
jrood-nrel: If it doesn't need an entirely new jacobian it should just be a runtime option.

malihass:
> Very cool! Can we try a performance test and see if there is any benefit?

Yes @esclapez is running some performance test with LMeX. BTW, I think that we should not merge until we get improved performance

gardner48:
> 2\. @gardner48 or @balos1 - can you tell us if the jacobian is stored as row or column major in sundials?

The sundials and magma dense matrices both use column major ordering.

hsitaram:
> > 2. @gardner48 or @balos1 - can you tell us if the jacobian is stored as row or column major in sundials?
> 
> The sundials and magma dense matrices both use column major ordering.

That changes everything, @malihass . Thank you David

malihass:
Closing since transposing the Jacobian does not change the strided access issue

balos1:
@malihass It turns out that it might be useful to have a row-major ordered Jacobian for use with the Ginkgo solvers. Do you still have this branch around to so I can we can try it?

EDIT: I am made this PR into a patch and I am trying it out with Ginkgo

malihass:
> @malihass It turns out that it might be useful to have a row-major ordered Jacobian for use with the Ginkgo solvers. Do you still have this branch around to so I can we can try it?
> 
> EDIT: I am made this PR into a patch and I am trying it out with Ginkgo

Ok, I restored the branch by the way (`malihass/rollJac`). Let me know if you need help with how to use the option

PR #358, opened 2023-02-17
Remove a bunch of pointers

jrood-nrel:
I can't take all the pointer dereferencing anymore.

### DIFF
--- a/Eos/Fuego.H
+++ b/Eos/Fuego.H
@@ -30,7 +30,7 @@ struct Fuego
   AMREX_GPU_HOST_DEVICE
   AMREX_FORCE_INLINE
   static void
-  TY2Cp(const amrex::Real T, amrex::Real Y[NUM_SPECIES], amrex::Real& Cp)
+  TY2Cp(const amrex::Real T, const amrex::Real Y[NUM_SPECIES], amrex::Real& Cp)
   {
     CKCPBS(T, Y, Cp);
   }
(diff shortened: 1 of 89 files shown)


esclapez:
That'll remove the native compatibility with a Chemkin-based code, but it might actually be already out the window, so that's probably fine.

marchdf:
Do we know of any chemkin-based codes that are hooking into pelephysics and using our API?

esclapez:
I don't know any ... I'm just mentioning it because that was the rational behind Fuego at the beginning, but now that we've moved away from it, it seems natural to go more modern/efficient implementation.

jrood-nrel:
If anybody has any objections, I will be fine not making the changes. These changes just felt like a natural progression. I have some ideas I'm exploring in making this stuff faster on the GPU, but those don't really need this.

marchdf:
I like this PR ;)

jrood-nrel:
Alright let's decide if we want to do this or not.

hsitaram (approved):
all good! Its high time we remove these remnants from the C -> fortran -> C++ builds from back in the day.

baperry2 (approved):
I like having all the consts where appropriate in the EOS functions to make it crystal clear what is an input/output.

Having directly chemkin compatible interfaces is cool in principal but if no one is actually using the capability there's no point in making the code worse to keep it around.

marchdf (approved):

malihass (commented):
This looks good to me. I ran a 0D reactor and compared successfully to Cantera for `dodecane_lu` and `dodecane_lu_qss`. Can we try to update the qss mechs as well?

malihass (approved):

esclapez (approved):

PR #359, opened 2023-02-17
Some cleanup for the mechanisms

marchdf:
One commit is for the ceptr changes, the other is the application to the mechs.

- [x] precompute imw
- [x] make some loops out of unrolled stuff

If this does nothing for performance, at least it reduces the numbers of LOC in the files... Maybe the most controversial part of this is the precomputing of imw...

### DIFF
--- a/Support/Mechanism/Models/Aromatic_KrNara/mechanism.cpp
+++ b/Support/Mechanism/Models/Aromatic_KrNara/mechanism.cpp
@@ -887,13 +887,12 @@ CKKFKR(
   amrex::Real q_f[],
   amrex::Real q_r[])
 {
-  int id;             // loop counter
   amrex::Real c[158]; // temporary storage
   amrex::Real PORT =
     1e6 * P / (8.31446261815324e+07 * T); // 1e6 * P/RT so c goes to SI units
 
   // Compute conversion, see Eq 10
-  for (id = 0; id < 158; ++id) {
+  for (int id = 0; id < 158; ++id) {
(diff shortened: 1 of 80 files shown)


baperry2 (approved):
Precomputing IMW isn't controversial to me. Seems like it would only be an issue if someone decided to run with more than double precision. 

Have you checked performance at all? Should show up with the ReactEval or even the EosEval case if there is any change

marchdf:
I haven't check performance yet... but yeah we should!

jrood-nrel:
I've been doing these changes manually and they appear to be just slightly faster.

esclapez (approved):
The inverse MW might change regtest to machine precision, but it is probably worth it !

marchdf:
yeah we do `get_imw` all the time..

malihass:
I do not see a speed up when running 0D reactors 20 times with `dodecane_lu` on my laptop
But this might be beneficial on other architectures
base
```
AJ
	 mean=2.5207885, std=0.031828763921176714
Direct
	 mean=3.879539473684211, std=0.050354672022165266
GMRES
	 mean=2.369563500000001, std=0.20342455422772837
```

cleanup
```
AJ
	 mean=2.5144515, std=0.06423357231814218
Direct
	 mean=4.142346999999999, std=0.1432226670991711
GMRES
	 mean=2.4172199999999995, std=0.23015836719528576 
```

malihass (approved):

jrood-nrel (approved):

PR #364, opened 2023-03-11
Setup CMake

esclapez:
to either compile PelePhysics as a stand-alone library or begin able to add PelePhysics to a project with:

<Pass PelePhysics options>
add_subdirectory(${PELE_PHYSICS_LOCATION})

target_link_libraries(myApp PUBLIC PelePhysics::${PELEPHYSICS_LIB}_api)

The lib name depends on the CHEM/EOS/TRANSPORT model it was compiled with.


esclapez:
@jrood-nrel I am fairly clueless when it comes to CMake, so hopefully this can still be useful. Tested the stand alone build of the library with CUDA and HIP successfully, and the include with PeleLMeX with plain MPI so far.

jrood-nrel:
I think this is a good idea. I can try to test it out.

CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
+cmake_minimum_required(VERSION 3.14)
```
jrood-nrel: I would set this to 3.23. Even though I think HIP support exists in lower versions, other projects are requiring 3.23 now.

CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+    HOMEPAGE_URL "https://github.com/AMReX-Combustion/PelePhysics"
+    LANGUAGES CXX
+    )
+
+message(STATUS "CMake version: ${CMAKE_VERSION}")
+
+#
+# Check if CMAKE_BUILD_TYPE is given. If not, use default
+#
+if ( NOT CMAKE_BUILD_TYPE )
```
jrood-nrel: I would remove this whole if statement. It's best to let CMake handle `CMAKE_BUILD_TYPE` itself.

CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+endif ()
+
+
+#
+# Define the object library to compile
+#
+set(pelephysics_lib_name PelePhysics-${PELEPHYSICS_CHEMISTRY_MODEL}-${PELEPHYSICS_EOS_MODEL}-${PELEPHYSICS_TRANSPORT_MODEL})
+add_library(${pelephysics_lib_name} OBJECT)
+if (BUILD_SHARED_LIBS)
+  set_target_properties(${pelephysics_lib_name} PROPERTIES POSITION_INDEPENDENT_CODE ON)
```
jrood-nrel: I never really know which way to go with this. Sometimes it's good to always have PIC on, but I think it's best to start by letting CMake handle it by removing this.

CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+
+endif ()
+
+#
+# Enable CUDA if requested
+#
+if (PELEPHYSICS_GPU_BACKEND STREQUAL "CUDA")
+   # CMake 3.18+: CMAKE_CUDA_ARCHITECTURES
+   # https://cmake.org/cmake/help/latest/policy/CMP0104.html
+   if(POLICY CMP0104)
```
jrood-nrel: I think we can remove this if setting the CMake minimum version higher.

CMakeLists.txt
```diff
@@ -0,0 +1,284 @@
...
+
+message( STATUS "PelePhysics configuration summary: ")
+message( STATUS "   Build type               = ${CMAKE_BUILD_TYPE}")
+message( STATUS "   Install directory        = ${CMAKE_INSTALL_PREFIX}")
+message( STATUS "   C++ compiler             = ${CMAKE_CXX_COMPILER}")
+message( STATUS "   C++ include paths        = ${_cxx_includes}")
+message( STATUS "   CMake target             = ${pelephysics_lib_name}_api")
+
+# Create non-object library for use as external target
+add_library(${pelephysics_lib_name}_api STATIC)
```
jrood-nrel: I wouldn't choose `STATIC` or `SHARED` here and let CMake decide.

jrood-nrel (approved):

PR #370, opened 2023-03-21
Calculate some CK routines without using thread private arrays

jrood-nrel:
@esclapez Curious if you could try this in PeleLMeX.


jrood-nrel:
@PaulMullowney This should give better GPU performance.

PaulMullowney:
> @PaulMullowney This should give better GPU performance.

Hmmm. I wonder if I can backport this into my old PelePhysics branch.

jrood-nrel:
I also inlined CKCVBS and removed thread private arrays, which is another main routine.

jrood-nrel:
These are my results on Crusher with PMF dodecane_lu:
Screen Shot 2023-03-21 at 1 00 36 PM

jrood-nrel:
Actually maybe these don't affect PeleLMeX since it uses a different `reactor_type`?

esclapez:
That's my guess, we might need to do that for the enthalpy routine too.

esclapez:
CKHBMS  and CKCPBS

jrood-nrel:
I can try that. Since I tried to generalize the thermo generation, I was able to do `CKCVBS` easily.

Support/ceptr/gen_qss.sh
malihass, line 1: Could we remove this file and use the conversion tool from CEPTR instead?

You should be able to regenerate the `qssa.yaml` with 
`poetry run qssa -lq ${PELE_PHYSICS_HOME}/Support/Mechanism/Models/list_qss_mech`

and then regenerate the cpp files with 
`poetry run convert -lq ${PELE_PHYSICS_HOME}/Support/Mechanism/Models/list_qss_mech`

jrood-nrel:
There are some other trivial routines that just do `get_imw` that can be inlined to remove arrays as well. This inlining really increases the code, but it looks beneficial for the GPU.

Support/ceptr/gen_qss.sh
jrood-nrel: Sure.

jrood-nrel:
I added CKHBMS and CKCPBS.

esclapez (approved):
A couple of seconds shaved off in LMeX too, nice !

PR #372, opened 2023-03-21
Compute typical values on device

jrood-nrel:
Optimizations from Paul.


jrood-nrel:
Screen Shot 2023-03-22 at 2 20 22 PM

esclapez:
Pretty nice improvement of the time spent in Cvode! That works for CYOrder but all the matrix-based approaches need a YCOrder. Can we make that a runtime internal parameter based on the type of linear solver selected for CVODE ?

jrood-nrel:
Oh yeah I forgot about that. We can probably do that, or somehow we could invert all the other accesses and use `YCOrder` somehow?

esclapez:
I think you need the `CYOrder` for coalescing memory access in the kernels, so it's probably part of the gain you get there when using plain GMRES.

Reactions/ReactorBase.H
```diff
@@ -88,11 +89,18 @@ public:
...
+#if defined(AMREX_USE_HIP) || defined(AMREX_USE_CUDA)
+  ~ReactorBase() { amrex::The_Arena()->free(m_typ_vals_gpu); }
+#else
   ~ReactorBase() override = default;
+#endif
 
 protected:
   int verbose{0};
   amrex::Array<amrex::Real, NUM_SPECIES + 1> m_typ_vals = {0.0};
+#if defined(AMREX_USE_HIP) || defined(AMREX_USE_CUDA)
```
marchdf: So why don't we make `m_typ_vals` a `GpuArray` (since it is a fixed size array) and then we just carry that around instead of also carrying a `_gpu` version around as well? Then we don't need the alloc/free/copy from host? I must be missing something...

jrood-nrel: I think you might be right. I haven't really looked into how the typical values are recomputed if so.

marchdf: Ok I think I did that... hopefully this works.

Reactions/ReactorCvode.H
```diff
@@ -51,7 +51,11 @@ class ReactorCvode : public ReactorBase::Register<ReactorCvode>
 public:
   static std::string identifier() { return "ReactorCvode"; }
 
+#if defined(AMREX_USE_GPU) && !defined(PELE_USE_MAGMA)
```
esclapez: MAGMA is not the only matrix-based approach we have on GPU. Having this a runtime parameter based on the choice of linear_solve is not possible ?

marchdf: ooof. That is not what I envisioned when doing this templating way back in the day... I will need to think about that. I guess we need to figure out what the best default is for now?

esclapez: Maybe we can switch based on a PeleC vs PeleLM basis if we have a define that can handle that.

Reactions/ReactorBase.cpp
```diff
@@ -96,29 +97,51 @@ ReactorBase::set_sundials_solver_tols(
...
-        ratol[offset + k] = m_typ_vals[k] * absTol;
+#if defined(AMREX_USE_HIP) || defined(AMREX_USE_CUDA)
+    const int nbThreads = 256;
+    const int nbBlocks = std::max(1, neq_tot / nbThreads);
+    auto arr = m_typ_vals.data();
+    amrex::launch_global<256>
+      <<<nbBlocks, 256>>>([=] AMREX_GPU_DEVICE() noexcept {
+        const int icell = blockDim.x * blockIdx.x + threadIdx.x;
+        if (icell < neq_tot) {
+          ratol[icell] = arr[icell / ncells] * absTol;
```
esclapez: This loop should depend on the `Ordering` no ? Instead of using `[icell/ncell]` to access arr, using `icell%(NUM_SPECIES+1)`

Reactions/ReactorBase.cpp
```diff
@@ -96,29 +97,51 @@ ReactorBase::set_sundials_solver_tols(
...
                      << " atolfact = " << absTol << " in PelePhysics \n";
     }
-    for (int i = 0; i < ncells; i++) {
-      const int offset = i * (NUM_SPECIES + 1);
-      for (int k = 0; k < NUM_SPECIES + 1; k++) {
-        ratol[offset + k] = m_typ_vals[k] * absTol;
+#if defined(AMREX_USE_HIP) || defined(AMREX_USE_CUDA)
+    const int nbThreads = 256;
+    const int nbBlocks = std::max(1, neq_tot / nbThreads);
+    auto arr = m_typ_vals.data();
```
esclapez: No need for the `.data()`, the GpuArray can be captured by the lambda directly.

marchdf: great point. fixed

Reactions/ReactorBase.cpp
marchdf: Ok so I fixed the second point (using size of `arr` though)`. But you bring up a good point about the ordering... That still needs to be fixed

esclapez: Yes. What was there before your commit was good for CYOrder, this works for YCOrder.

marchdf: Yeah it means moving a bunch of stuff around and templating this function ;) working on it

marchdf: Ok so I think I managed it... the big difference I had to make with what Paul/Jon had is now the kernel is parallelized on ncells with a for loop on the elements of `typ_vals` and not each element in `ratol`. That probably has a performance impact. I will take suggestions...

marchdf:
Does anyone have a good test case lying around to make sure we didn't mess anything up in this PR?

Reactions/ReactorCvode.H
marchdf: Haven't figured this out yet... One solution I have right now is an if condition around each call where there is an `<ordering>` template argument. So we would have if condtions inside kernel calls. And a whole bunch of 
```
if (ordering == ycorder){
   utils::box_flatten<YCOrder>( icell, i, j, k, ncells, captured_reactor_type, captured_clean_init_massfrac, rY_in, rYsrc_in, T_in, rEner_in, rEner_src_in, yvec_d, udata_g->rYsrc_ext, udata_g->rhoe_init, udata_g->rhoesrc_ext);
}
else{
  utils::box_flatten<CYOrder>( icell, i, j, k, ncells, captured_reactor_type, captured_clean_init_massfrac, rY_in, rYsrc_in, T_in, rEner_in, rEner_src_in, yvec_d, udata_g->rYsrc_ext, udata_g->rhoe_init, udata_g->rhoesrc_ext);
}
```
Not too crazy about this idea but I could be convinced for performance reasons. I think it would lead to larger executables.

Another idea is that we make it a compile flag? Instead of switching on GPU or LM/C we just are explicit about the one we want? We could default it based on other stuff like GPU and LM/C...

jrood-nrel:
```
pmf-dodecane.inp geometry.prob_lo=0.0 0.0 0.0 geometry.prob_hi=3.2 3.2 1.6 amr.n_cell=256 256 128 prob.L=0.4 0.4 1.6 max_step=10 amr.plot_files_output=0 amr.plot_int=10 amr.checkpoint_files_output=0 amrex.abort_on_out_of_gpu_memory=1 pelec.cfl=0.1 pelec.init_shrink=1.0 pelec.change_max=1.0 amrex.the_arena_is_managed=0 pelec.chem_integrator=ReactorCvode cvode.solve_type=GMRES amr.blocking_factor=16 amr.max_grid_size=64 pelec.use_typ_vals_chem=1 ode.rtol=1e-4 ode.atol=1e-5 pelec.typical_rhoY_val_min=1e-6 amrex.use_gpu_aware_mpi=1 amrex.the_arena_init_size=17179869184
```

esclapez:
I'll give that a try in LMeX

marchdf:
Thanks! I don't get diffs on PeleC...

Reactions/ReactorCvode.H
esclapez: Adding yet another compile flag sort of goes against adding runtime flexibility. I think we can probably make the `if` work but maybe have the if outside of the kernels if possible. Let me have a look.

marchdf: Should we merge this PR and then do that? My guess is that it's going to be some amount of work...

esclapez: In the current state, we break using GMRES if you compile with MAGMA ON for example (assuming we have typical values ON in both cases).

esclapez: Currently, we break the `sparse_direct` solve on CUDA GPUs. We have some collaborator using this option.

marchdf: gotcha. How about we remove that change (which is not about typ vals anyway) from this PR, open a new one with the change and then the stuff to figure out how to runtime it? Should I do that?

esclapez: That sounds good to me.

esclapez:
No difference on LMeX either and performance-wise it's pretty close to what I got with the initial implementation without the Ordering templating.

Reactions/ReactorCvode.H
marchdf: Ok done in #383

marchdf (approved):

esclapez (approved):

PR #375, opened 2023-03-31
Put molecular weights in global array

jrood-nrel:


esclapez (approved):
Not as much gain in LMeX using dodecane_lu_qss as in PeleC, but still something !

PR #379, opened 2023-04-05
Update SUNDIALs and MAGMA

jrood-nrel:
I haven't tried this yet. I think since Frontier is live now, I decided to remove the local file search.


esclapez (approved):
Nice !

PR #381, opened 2023-04-17
Redoing global (i)mw again

marchdf:
So we could do this to get around the `global_(i)mw` calls on device and host. Basically we detect where the function is being executed with `AMREX_DEVICE_COMPILE`

What do we think?


baperry2 (approved):
I don't mind this at all in the machine generated mechanism code. But in the Eos and Reactor code it does reduce readability significantly.

marchdf:
yes it's unfortunate... Though your point made me think that I was being dumb. I could make a function that is 

```
AMREX_GPU_HOST_DEVICE
  AMREX_FORCE_INLINE
  amrex::Real mw(const int n){
#if AMREX_DEVICE_COMPILE
        return global_mw[n]
#else
        return h_global_mw[n]
#endif
}
```
and then remove all my `ifdef`s and replace by that function call. That would have been much better. I will do that.

PR #385, opened 2023-04-20
Better global mw

marchdf:
Replaces #381


jrood-nrel:
I'm interested if this has a performance penalty.

baperry2 (approved):

marchdf:
I would assume it would be the same performance, it should be inlined and the compiler will only see one of those branches.

esclapez (approved):
Just tested against development in LMeX, showing fairly marginal gain (0.2 %), but similar to the previous implementation, so I think it is good.

jrood-nrel (approved):

PR #397, opened 2023-05-10
Restore CVODE OMP support

esclapez:
Move the CVODE-related objects (solution vector, Jac. matrix, linear solver, ...) out of the Reactor class, they are now created upon the react() call as in the GPU case. No significant performances hit in the non-threaded version of the code.


marchdf (approved):

jrood-nrel (approved):
This shows no data races in PeleC PMF.

PR #405, opened 2023-08-21
Fully GPU enabled set_sundials_solver_tols

cbauinge:
The SYCL computations in set_sundials_solver_tols were still performed on the CPU, resulting in a big bottleneck. I implemented them on the GPU.

### DIFF
--- a/Reactions/ReactorUtils.H
+++ b/Reactions/ReactorUtils.H
@@ -111,7 +111,7 @@ set_sundials_solver_tols(
     neq_tot, /*use_managed_mem=*/false,
     *amrex::sundials::The_SUNMemory_Helper(),
     &amrex::Gpu::Device::streamQueue(), sunctx);
-    amrex::Real* ratol = N_VGetDeviceArrayPointer_Sycl(atol);
+  amrex::Real* ratol = N_VGetDeviceArrayPointer_Sycl(atol);
 #else
   N_Vector atol = N_VNew_Serial(neq_tot, sunctx);
   amrex::Real* ratol = N_VGetArrayPointer(atol);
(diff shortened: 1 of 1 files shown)


jrood-nrel:
@cbauinge Thanks for this. We just require `clang-format -i` to be run on the source files for it to get merged.

cbauinge:
> @cbauinge Thanks for this. We just require `clang-format -i` to be run on the source files for it to get merged.

Hi. Sorry about that. I did it now. Please let me know if there is something missing.

jrood-nrel:
It looks like this doesn't work on the CPU. Instead, would we be able to change:
```
#if defined(AMREX_USE_HIP) || defined(AMREX_USE_CUDA)
```
to
```
#if defined(AMREX_USE_GPU)
```
and expect this to work with SyCL as well?

jrood-nrel:
Or we could just have an extra block with this code with:
```
#elif defined(AMREX_USE_SYCL)
```

jrood-nrel:
I'm seeing errors when running this on Intel GPUs as well:
```
... Initializing reactions, using interval dt = 1.498288935e-08

[CVODE ERROR]  CVodeSVtolerances
  abstol has negative component(s) (illegal).

amrex::Abort::7::Problem in setSundialsSolverTols !!!
```
I made another PR to deal with SyCL separately https://github.com/AMReX-Combustion/PelePhysics/pull/406

jrood-nrel:
Actually I think all these code blocks can be a 1D `ParallelFor`. I will try that.

jrood-nrel:
I think https://github.com/AMReX-Combustion/PelePhysics/pull/406 solves all these code blocks and the performance is similar (on Intel GPUs so far). So I think I will close this and use that PR.

cbauinge:
Sorry for the delayed response. Looks like you managed to solve it, thank you!

PR #406, opened 2023-08-22
Compute cvode typical values using ParallelFor

jrood-nrel:


jrood-nrel:
Screen Shot 2023-08-22 at 12 18 24 PM
Performance is slightly better than using the SyCL code block on Intel GPUs. I haven't tested AMD/Nvidia yet.

marchdf (approved):

PR #409, opened 2023-08-28
Upgrade CEPTR to cantera 3.0

### LINKED ISSUE #408: Enable CEPTR with Cantera version 3.0.0
A significant new version of Cantera has been released: https://cantera.org/documentation/release_notes/v3.0.0.html

This version has some changes under the hood that make it incompatible with the existing CEPTR code (reaction third body efficiencies moved to a different location, reaction types renamed, perhaps others)

For now, poetry will be set up for CEPTR to use Cantera version 2.6 (#407), but it would be good to stay up to date with Cantera.

marchdf:
To do:
- [x] no diffs when running `list_mech` conversion
- [x] no diffs when running `list_qss_mech` conversion 

Closes #408


baperry2:
Convert `reaction.efficiencies` -> `reaction.third_body.efficiencies` and `reaction.default_efficiency` -> `reaction.third_body.default_efficiency` to eliminate the deprecation warnings

Support/ceptr/ceptr/qssa_converter.py
```diff
@@ -1621,34 +1621,34 @@ def qssa_coeff_functions(
...
                 troe = reaction.rate.falloff_coeffs
                 ntroe = len(troe)
                 is_troe = True
-            elif reaction.rate.type == "Sri":
+            elif "Sri" in reaction.reaction_type:
                 pass
                 # sri = reaction.rate.falloff_coeffs
                 # nsri = len(sri)
                 # is_sri = True
-            elif reaction.rate.type == "Lindemann":
```
malihass, line 1651: Should we keep the `==` ? that seems dangerous to be lenient here
It seems to be`LindemannRate` `TroeRate`, `SriRate` and `TsangRate`

Support/ceptr/ceptr/qssa_converter.py
```diff
@@ -2167,35 +2167,35 @@ def qssa_component_functions(
...
                 # troe = reaction.rate.falloff_coeffs
                 # ntroe = len(troe)
                 # is_troe = True
-            elif reaction.rate.type == "Sri":
+            elif "Sri" in reaction.reaction_type:
                 pass
                 # sri = reaction.rate.falloff_coeffs
                 # nsri = len(sri)
                 # is_sri = True
-            elif reaction.rate.type == "Lindemann":
```
malihass, line 2198: Same comment, I think it would be sager to keep an `==`

Support/ceptr/ceptr/reaction_info.py
```diff
@@ -74,7 +71,7 @@ def sort_reactions(mechanism):
     # simplest case
     for k, r in enumerate(reaction_info.rs_unsorted):
         if r not in reaction_info.rs:
-            if r.reaction_type != "three-body":
```
malihass, line 77: Same comment, except if there is a good reason to not do an exact match

Support/ceptr/ceptr/qssa_converter.py
marchdf: Good point. Looks like there is now 
```
In [7]: reac.rate
Out[7]: <TroeRate at 1066df870>

In [8]: reac.rate.type
Out[8]: 'falloff'

In [9]: reac.rate.sub_type
Out[9]: 'Troe'
```
so I will switch to using those tomorrow instead of using
```
In [10]: reac.reaction_type
Out[10]: 'falloff-Troe'
```

malihass:
For the qss test, are you trying with or without AJ, or both?

marchdf:
For the qss, I run the `convert -lq` on the `list_qss_mech` file. Is that ok?

baperry2 (approved):

Support/ceptr/ceptr/qssa_converter.py
marchdf: done

Support/ceptr/ceptr/reaction_info.py
marchdf: done

marchdf:
I apologize for the noise... I had to make some updates to black for fix some string formatting I introduced and that made a bunch more changes that I was not intending.

malihass:
> For the qss, I run the `convert -lq` on the `list_qss_mech` file. Is that ok?

That works ! This is with Jacobian

malihass (approved):

PR #410, opened 2023-08-28
Use the sundials submodule for make TPL

jrood-nrel:


marchdf (approved):

baperry2 (approved):
Just a note that users compiling the PelePhysics tests with GNUmake after this PR will need to clone their own copy of Sundials and specify the path using the `SUNDIALS_HOME` in the `GNUmakefile` for the case (in addition to cloning AMReX and setting `AMREX_HOME`).

We should really have that documented somewhere, but don't really have anything on these test cases, so that's a much bigger task beyond the scope of this PR.

jrood-nrel:
We could check if `SUNDIALS_HOME` isn't set and then do the wget if necessary.

baperry2 (approved):
I like the idea of using wget as a back up, but what's there now is also fine.

Issue #162, opened 2021-06-14
Fuego w/ FORD rates generating incorrect prefactor conversion term


whitmanscu:
With the Fuego-generated mechanisms for kinetics with FORD rates, species production/consumption rates appear unphysically low. I believe this is due to the generated rates in comp_qfqr, where the prefactor conversion appears to not be properly implemented for FORD rates. There is conversion that happens between cgs and mks before productionRate is called, and the prefactor term seems to be intended to convert back the other way based on the mechanism units. However, the `dim` that is used in calling `self._prefactorUnits` in `CPickler.py` for the prefactor term is based off of the stoichiometric coefficients rather than the FORD rates, which leads to an improper conversion. If I manually adjust the prefactor terms based off of the FORD rates, my kinetics appear to give realistic production/consumption behavior.

A quick fix seems like it would be to change
https://github.com/AMReX-Combustion/PelePhysics/blob/283a0475e2487ed570eb47c3118d0fc981e40e05/Support/Fuego/Pythia/pythia-0.4/packages/fuego/fuego/serialization/c/CPickler.py#L1751
to something like
```
if (len(reaction.ford) > 0) :                
   dim = self._phaseSpaceUnits(reaction.ford)            
else:                
   dim = self._phaseSpaceUnits(reaction.reactants)
``` 
similar to what's done in the initialization. However, I want to check and make sure this is the most appropriate fix, or if a different route might be better.

EnnaDelfen:
Sounds like this is what is going on, and the fix proposed would solve the issue

drummerdoc:
@EnnaDelfen Is there any additional code that would impact the analytic Jacobian that would also need fixing?

EnnaDelfen:
I'm looking through it because he's specifically taking about productionRate_GPU I'm looking through to see if other functions are concerned

EnnaDelfen:
Yeah so I'm thinking there are a bunch of other places where the suggested modifications need to be implemented. It's tricky to catch them all without a mechanism/test case to test the implementations. Basically wherever a dim needs to be computed or a call to _QSSsortedPhaseSpace is needed, FORD need to be taken into account. I think no one has actually ued a FORD reaction scheme in where AJ or other fancy chemistry treatment was required and that's why it simply hasn't been implemented. SHould be a matter of grepping the code for calls to _QSSsortedPhaseSpace and dim and checking results. I'm happy to review the proposed changes !

drummerdoc:
[...]

Issue #23, opened 2019-06-06
Add FORD keyword handling in Fuego for analytic and approx Jacobian


drummerdoc:
A recent fix will (just did) add the native handling of the FORD keyword in order to properly deal with non-integer explicitly set reaction orders (such as what is used in the CH4 2-step mech).  This mod is not yet pulled through the code that generates Jacobians though.

In fact, there is generalized ability in place that for specific mechanisms, auxiliary files can be used to implement specific functions.  Typically this is used to implement a QSS model for CKWYP, for example.  However, those functions are not necessary called by the RHS functions in the ODE integrators, AND the are not consistent with the Jacobian code.  There should be some kind of indicator that only some of the functionality is available in these cases (until the full functionality is eventually added).

whitmanscu:
A recent fix (Issue #41) added some FORD keyword handling to the CPickler for Fuego, but it appears to remain incompletely implemented. The most immediate issue I see is that only the GPU version of comp_qfqr was modified to use the FORD orders while the CPU code still needs the same change.

I'm commenting on this issue because I'd like FORD rates to be fully implemented and it's not immediately clear to me what changes past what I mentioned above need to be made to the Jacobian generation (or other routines) in order to do this.

whitmanscu:
It looks to me like the code needs to be fixed for aJacobian_precond, aJacobian, and vcomp_wdot, in addition to the comp_qfqr CPU method I mentioned above. Am I missing anything? I tried making modifications in these places but I'm still getting crashes when I try and run a 2-step propane mechanism with FORD rates (I posted the chemistry file to the other issue).

malihass:
closed by #414

PR #414, opened 2023-09-07
fix rescaling for FORD

### LINKED ISSUE #318: Unit mismatch for elementary reactions with arbitrary orders of reaction
Hi, 

I was looking at some mechanisms which have reactions with arbitrary reaction orders and found that the units of the pre-exponential coefficient in the Arrhenius rate reactions are not the same as those given by Cantera. I added a few print statements (after line 196 in `production.py`) to compare the units given by Cantera with those computed by PelePhysics, and ran a test for the 3 reaction chem-CH4-2step mechanism (all the reactions prescribe arbitrary reaction orders).

The output of the run is in this file: [ceptr_output.txt](https://github.com/AMReX-Combustion/PelePhysics/files/9425718/ceptr_output.txt). The units are significantly off.

The fix may be fairly easy though. Instead of using 
`ctuc = cu.prefactor_units(cc.ureg("kmol/m**3"), 1 - dim)` as in line 196 of `production.py`, we could extract the unit from Cantera as  `ctuc = cc.ureg(str(reaction.rate_coeff_units).split("at")[0][7:-2].replace("^", "**"))`. Of course there may be a simpler (or cleaner way) of doing this without having to do these string operations.

I did try to test the original and the fix for the homogeneous reactor test using CeptrTest, but, somehow, I was not able to get the PelePhysics simulation to work for either case. If I could get some pointers on how to create the homogeneous reactor test with PelePhysics, I'd be happy to attempt a fix for this issue.

Cheers!

malihass:
closes #318 #162 #23 

- [x] verify that no mech is changed (including QSS)
- [x] Compare all FORD mechs to cantera
- [x] Update mechs
- [x] Revert the safeguard against FORD

Co-authored-by: Anirudh Jonnalagadda <jonnalagadda.anirudh@gmail.com>
Co-authored-by: Samuel Whitman

### DIFF
--- a/Support/Mechanism/Models/chem-CH4-2step/mechanism.H
+++ b/Support/Mechanism/Models/chem-CH4-2step/mechanism.H
@@ -2159,7 +2159,7 @@ aJacobian_precond(
   // reaction 0: 2 CH4 + 3 O2 => 2 CO + 4 H2O
   // a non-third-body and non-pressure-fall-off reaction
   // forward
-  phi_f = pow(sc[0], 3.000000) * (sc[2] * sc[2]);
+  phi_f = sc[0] * sc[2];
   k_f = 154500000 * exp(0.5 * tc[0] - (20075.8288822793) * invT);
   dlnkfdT = 0.5 * invT + (20075.8288822793) * invT2;
   // rate of progress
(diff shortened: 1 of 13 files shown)


malihass:
FORD mech comparisons
FD is for finite difference Jacobian
AJ is for analytical Jacobian
Figure_1
Figure_2

marchdf (approved):

Support/Mechanism/Models/chem-CH4-2step/mechanism.H
```diff
@@ -0,0 +1,2978 @@
...
+  // rate of progress
+  q = k_f * phi_f;
+  dqdT = dlnkfdT * k_f * phi_f;
+  // update wdot
+  wdot[0] -= 3 * q; // O2
+  wdot[1] += 4 * q; // H2O
+  wdot[2] -= 2 * q; // CH4
+  wdot[3] += 2 * q; // CO
+  // d()/d[O2]
+  dqdci = +k_f * 3.000000 * sc[0] * sc[0] * sc[2] * sc[2];
```
baperry2: I think this still needs to be updated for the FORD stuff. This appears to have been calculated using the stoichiometric coefficients as reaction orders rather than using the FORD reaction orders.

malihass: @baperry2 Good catch! I think I got all of it but I could use another pair of eyes to make sure I caught everything

baperry2 (approved):
Nice, looks good to me!

marchdf (approved):

Issue #323, opened 2022-10-02
Unit mismatch for three-body reactions


jAnirudh:
Hi,

I am not sure but I may have found a bug in the reaction rate coefficients units of three body reactions.
 
I was looking at the Li-Dryer Hydrogen combustion mechanism which has 4 three-body reactions and found that CEPTR adds an additional _meter ** 3 / kilomole_ factor to each three-body reaction when compared to Cantera.

I've made some basic [changes](https://github.com/AMReX-Combustion/PelePhysics/commit/306ab2ba4d760687cfce9c3243ad4e66aeb61dc7) to `production.py` and wrote a script `Testing/Exec/3bodyTest/threebody_Aunits.py` to get this output: [units_dump.txt](https://github.com/AMReX-Combustion/PelePhysics/files/9692099/units_dump.txt). Essentially this difference in units increases the magnitudes of the forward reaction rates by a factor of 1000.

I tried to see the effect of this by looking at the pressure scaling results reported [here in (Fig8d) ](https://arc.aiaa.org/doi/10.2514/6.2021-3448) but this additional 1000 multiplying factor does not seem to affect the results as shown below.

![1bar](https://user-images.githubusercontent.com/7749537/193446812-ffb0835d-4ae3-490d-976c-0461802cea75.png)
![5bar](https://user-images.githubusercontent.com/7749537/193446844-185b7b42-65ba-4a47-83a1-7411c1a14209.png)
![10bar](https://user-images.githubusercontent.com/7749537/193446849-a5099fea-2f2b-47e0-91bf-46c553c7fab8.png)

The results I've shown here can be reproduced from `Testing/Exec/3bodyTest/` in [this fork](https://github.com/jAnirudh/PelePhysics/tree/3BodyRxnUnits). For `unit_dump.txt` run `python threebody_Aunits.py` and for the remaining plots run `sh exec.sh`. 

I also tried computing the ignition delay for this problem but it turns out that the temperature profile does not show any deviation from the expected profiles.

Again, I am not sure if this a bug, but thought it may be worth reporting.

Cheers,
Anirudh

malihass:
Thanks for reporting this,  I finally got the chance to dive into this.

[...]

PR #416, opened 2023-09-08
Used Cantera reaction rate coefficient units

jAnirudh:
Proposed fix to #318 for computing the production rates for elementary reactions with FORD reactions

The reaction rate coefficients appearing as `ctuc` in `ceptr/production.py` are computed directly from Cantera, except for Three-Body reactions, which are made to work according to Cantera's legacy behaviour as discussed in #323.

These changes are reflected in the `comp_qfqr`, `productionRate` and the `productionRate_Light` methods of the corresponding `mechanism.H` files.

This fix should immediately work for FORD reactions when not using the analytic Jacobian.

### DIFF
--- a/Support/ceptr/ceptr/production.py
+++ b/Support/ceptr/ceptr/production.py
@@ -192,9 +192,11 @@ def production_rate(
             is_sri = reaction.rate.sub_type == "Sri"
             is_lindemann = reaction.rate.sub_type == "Lindemann"
             aeuc = cu.activation_energy_units()
+
+            ctuc = cc.ureg(str(reaction.rate_coeff_units).replace("^", "**"))
+
             if not third_body and not falloff:
                 # Case 3 !PD, !TB
-                ctuc = cu.prefactor_units(cc.ureg("kmol/m**3"), 1 - dim)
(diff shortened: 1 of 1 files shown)


malihass:
Thanks for the PR!
So instead of doing this, #414 adjusts the value of `dim` which is a little bit cleaner than the string operations.
Would you please be able to try and see if #414 fixes that issue for you?

jAnirudh:
No worries :)

I am not a fan of the string operations myself honestly. Ideally I'd have liked Cantera to simply give me the rate coefficient unit directly without me having to do any string manipulations. Maybe this is something the Cantera folks may like to consider for us.

However, I found this way to be much easier to follow and keep track of than modifying `dim` when adding new reactions. Here, irrespective of the kind of reaction, if Cantera supports it we're sure to get the same dim. This makes it much easier to implement new reactions plus there are fewer and more localized changes to track which is good for maintainability IMO.

Anyway, while I haven't been able to go through all the code changes in `production.py` and `util.py`,  this change and #414 give identical `mechanism.H` files for the 3 methods I've mentioned. So we essentially have 2 ways to solve the problem :)

malihass:
OK, I am going to close this since this results in the same mechanisms. Thanks for bringing these FORD issues to our attention!

PR #430, opened 2023-10-04
Add ifdef to remove BDF for large mechanisms on the GPU

jrood-nrel:


marchdf (approved):

Reactions/ReactorBDF.cpp
```diff
@@ -139,6 +139,7 @@ ReactorBDF::react(
 #endif
 )
 {
+#if !defined(AMREX_USE_GPU) || (NUM_SPECIES < 50)
```
PaulMullowney: dodecane_lu compiles so maybe make it 54?

PaulMullowney:
Do we need to be a little more careful with this? Doesn't BDF build on Nvidia with large chemistry models?

jrood-nrel:
I'm waiting for machines to come out of maintenance and then I will test this.

hsitaram:
I had checked builds and runs on nvidia gpus but not with a super large chemistry. i wonder if doing GpuArray instead of static arrays will fix the problem. ReactorCvode does gpuarrays as well

jrood-nrel:
This solves the issue for me. I'm going to merge this because it currently fixes something that is broken anyway.

PR #431, opened 2023-10-18
Update isooctane_lu mechanism to allow it to be generated

jrood-nrel:


baperry2:
While we're in here can you also delete some of the other files that are old/extraneous?

```
README_bigMech.md
sk143/PMF
COFD.dat
COFLAM.dat
COFETA.dat
COFTD.dat
```

I honestly have no idea what the `COF*.dat` files are, which makes me inclined to get rid of them, unless @drummerdoc objects.

baperry2 (approved):

jrood-nrel:
Also it looks like black updated or something, causing formatting failures?

baperry2:
Yeah, I guess so. I thought this would be easily fixed just by running `poetry run black .` in `Support/ceptr/ceptr` but it looks like the updated formatting from black conflicts with what flake8 expects so if you do that you then get errors on running  `poetry run flake8 .`. So IDK. This should be safe to force push, but we'll have to deal with that eventually.

drummerdoc:
> While we're in here can you also delete some of the other files that are old/extraneous?
> 
> ```
> README_bigMech.md
> sk143/PMF
> COFD.dat
> COFLAM.dat
> COFETA.dat
> COFTD.dat
> ```
> 
> I honestly have no idea what the `COF*.dat` files are, which makes me inclined to get rid of them, unless @drummerdoc objects.

Back in prehistoric times, the data in these files was generated by the convert function that made EGLIB init data from CHEMKIN inputs.  That data was originally generated by CHEMKIN's function tranfit and dumped into a "Linkeg" file that EGLIB init expected to be in the folder where the codes were run that called EGLIB functions. I decided to instead write that data into arrays, and appended the setting of those arrays into functions called in my own version of the eginit function.  Since then, someone decided to dump that data into separate files...I have no clue why or what has become of that strategy, but I'm reasonably sure that data is still ultimately required for the EGLIB functionality we use under the covers.

PR #433, opened 2023-10-23
Clip the concentrations on fractional pows

marchdf:
When doing fractional powers, make sure the concentrations are not negative.

Todo
- [x] regen qss mechs
- [x] remove the fixme on the sqrt (what's up with that @malihass?)

This is related to https://github.com/AMReX-Combustion/PeleLMeX/discussions/287

### DIFF
--- a/Support/Mechanism/Models/JL4/mechanism.H
+++ b/Support/Mechanism/Models/JL4/mechanism.H
@@ -1913,15 +1913,17 @@ comp_qfqr(
 {
 
   // reaction 0: 2 CH4 + O2 => 2 CO + 4 H2
-  qf[0] = pow(sc[0], 0.500000) * pow(sc[1], 1.250000);
+  qf[0] =
+    std::sqrt(std::max(sc[0], 1e-14)) * pow(std::max(sc[1], 1e-14), 1.250000);
   qr[0] = 0.0;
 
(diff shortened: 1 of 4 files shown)


baperry2:
For computing reaction rates, I'd favor clipping to zero rather than positive epsilon, which should still always be valid as long as the FORD orders are positive (this is true for the FORD mechs we currently have but IDK if it's universal). It's hard to know exactly what order of magnitude is "small enough" for these, especially with the fractional exponents (1e-12 ^ 0.25 = 1e-3). I think this might be why the case is still failing here: https://github.com/AMReX-Combustion/PeleLMeX/discussions/287#discussioncomment-7363458

For the Jacobian, we can get negative exponents so we need to clip to epsilon rather than 0, but that is ok.  Making a slight error in the Jacobian could impact rate of convergence, but should still allow us to converge to the correct solution within a specified tolerance.

marchdf:
How about clip to 0 if order > 0, clip to 1e-14 if order < 0 ?

baperry2:
> How about clip to 0 if order > 0, clip to 1e-14 if order < 0 ?

Sounds good to me.

marchdf:
Ok I gave that a shot.

baperry2 (approved):

PR #434, opened 2023-10-25
enable COMP=clang for TPL

baperry2:
The AMReX GNUmake system interprets `COMP=llvm` and `COMP=clang` equivalently, so we should too for `TPL`.


jrood-nrel:
While you're at it, can you change `intel` to `intel-classic`?

jrood-nrel (approved):

PR #435, opened 2023-10-26
 Remove tc thread private array

marchdf:


marchdf:
@malihass can you check to see if this gives correct results? I think it does but your tests are better ;)

jrood-nrel:
This appears to have no performance implications.

jrood-nrel (approved):

malihass:
I just tested 0D ignition on `ch4_2steps`, `dodecane`, `dodecane_lu_qss` with and without AJ and I do not see significant differences in the temperature profile

malihass:
I see a very small diff of temperature specifically when running the 0D ignition with the analytical Jacobian and the `dodecane_lu_qss`. I don't see any diff for `dodecane_lu` and `ch4-2-steps`. I also don't see diff for the qss mech with numerical Jacobian. Maybe something is off with the symbolic Jacobian?

Otherwise for performance, I suspect it will shine most on GPU. Did you see no diff when running on GPU @jrood-nrel ?

marchdf:
I don't expect a performance change (it's basically what I want). I am not really doing this for performance... but i am hoping this gets me to the nasa9 polynomials easier

marchdf:
I think there are a couple of things causing this minor diff in the analytical jacobian:
- `pow(x, 0.5)` became `sqrt(x)`
- `logT`, `T2`, etc are expressions instead of using `tc[i]`. 
All this might lead to a slightly different expression optimization path.

Thoughts?

malihass (approved):
I don't see an error in the ceptr code. The changes also do modify a bit the common sub expressions being identified (increases by 0.1% in the case of dodecane) which could explain the very small diff.
I went ahead and compared the actual Jacobians on 100 random samples of composition space and they are the same up to machine precision (I did this for `CH4_lean_qss`, `dodecane_lu_qss` and `heptane_lu_qss`). So I am convinced! Thanks for bearing with me!

PR #436, opened 2023-10-26
diagPDF: barriers on atomic add to bins

baperry2:
I believe this will fix the sporadic errors we've been getting on LMeX tests.


marchdf (approved):

baperry2:
From the results of https://github.com/AMReX-Combustion/PeleLMeX/pull/293 it looks like this worked.

PR #437, opened 2023-10-27
Add multiprocessing for ceptr

marchdf:


jrood-nrel:
Could you just do python multithreading rather than needing an MPI dependency?

marchdf:
That's a better idea. Let me try

marchdf:
Ok that was fun. 

Generating all the normal mechs goes from 72s to 16s. Generating all the qss mechs goes from 1400s to 460s.

jrood-nrel (approved):

PR #438, opened 2023-10-30
Simplify ceptr python code

marchdf:
- simplify a bunch of code
- remove `sys.exit` and replace with `raise` error. This is because the use of the `multiprocessing` (PR #437) does not like child processes just exiting via sys.exit. It can basically cause a hang in the main process. Also `raise` errors is just a better thing to be doing
- no diffs caused in the mech files


jrood-nrel (approved):

malihass (approved):
Thanks for fixing this!

PR #439, opened 2023-10-30
Move dcvpdtemp

marchdf:


baperry2 (approved):

malihass (approved):
If you have a way to move the dcvpRdT call without it showing changes in the mechs, that would be great! Otherwise it is ok as is!

marchdf:
Yeah I wish I could... I tried but couldn't make it happen.

PR #443, opened 2023-11-30
Add the Util folder from PeleMP

marchdf:
What do we think? I think this preserves (some) history (at least authorship) and I can do things piece by piece.


jrood-nrel:
I'm confused by this so far, but it's still a draft.

marchdf:
This starts the migration of files from PeleMP to PelePhysics. I started with files that matter less.

baperry2:
I like that authorship is preserved and things can be done incrementally, so I think this is a good approach for bringing stuff in from PeleMP.

I do think we should consider how best to organize things as we move stuff from PeleMP though. Right now we have a mix of both C++ source code and python utilities in both the Support and Utility, which I think is a bit confusing.

jrood-nrel (approved):

PR #445, opened 2023-11-30
Filter Utility

baperry2:
Brings the Filter functionality from PeleC to make it accessible for other Pele codes.


marchdf (approved):

Utility/Filter/Filter.cpp
```diff
@@ -39,7 +39,7 @@ Filter::set_gaussian_weights()
     // equivalent to: std::sqrt(gamma / (PI * _fgr * _fgr)) * std::exp((-gamma
     // * (i-_ngrow) * (i-_ngrow)) / (_fgr * _fgr));
     _weights[i] =
-      1.0 / (std::sqrt(2.0 * constants::PI()) * sigma) *
+      1.0 / (std::sqrt(2.0 * amrex::Math::pi<amrex::Real>()) * sigma) *
```
marchdf, line 42: well well. Maybe we should pull that in everywhere? or alias `constants::PI` to amrex pi?

baperry2: I'll do that in the PeleC PR for this.

PR #448, opened 2023-12-05
Move PeleMP Source files to PelePhysics

marchdf:


baperry2:
Looks good, but I think we should wait to merge until opening parallel PRs in C and LMeX that account for the change and pass the relevant tests

jrood-nrel (approved):

jrood-nrel:
I don't mind merging any time. The other projects point to their respective commits of PelePhysics that work until we change them.

marchdf:
@baperry2  what Jon said was what I was thinking. I guess I was thinking we get PelePhysics to where we want it to be (for the MP stuff at least) and then update the submodules in the other codes and make the appropriate changes. Not set on this obv if you have other thoughts.

baperry2 (approved):
My thought was that it could get unwieldy to make all the corresponding changes in C and LMeX at once rather than incrementally. But up to you!

PR #449, opened 2023-12-06
Move PeleMP docs into PelePhysics

marchdf:


jrood-nrel (approved):

Docs/sphinx/Spray.rst
```diff
@@ -0,0 +1,490 @@
...
+
+4. If injection occurs, the amount of mass injected, :math:`m_{\rm{actual}}`, is summed and compared with the desired mass flow rate. If :math:`m_{\rm{actual}} / t_{\rm{inj}} - \dot{m}_{\rm{inj}} > 0.05 \dot{m}_{\rm{inj}}`, then :math:`N_{P,\min}` is increased by one to reduce the likelihood of over-injecting in the future. A balance is necessary: the higher the minimum number of parcels, the less likely to over-inject mass but the number of time steps between injections can potentially grow as well.
+
+Spray Validation
+================
+
+Single Droplet Tests
+--------------------
+
+Single droplet tests are performed and compared with computational or experimental results published in literature. These tests are setup in ``PeleProduction/PeleMPruns/single_drop_test``. To run a test case, simply open ``Validate.py`` and set the case name from the table below ::
```
baperry2, line 440: This is out of date and never made it into the main PeleProduction branch. But when we bring the cases over to PeleC and PeleLMeX, being able to reproduce these results would be a good objective.

baperry2 (approved):
I'm ok with merging this as is because we can't really fix the incorrect paths to cases until the cases get moved anyway.

Docs/sphinx/Spray.rst
marchdf: ah yes that's a good point and a good objective

PR #452, opened 2023-12-08
Move a bunch of directories around

marchdf:

PR #456, opened 2023-12-12
Fix error of PmfData on GPU

### LINKED ISSUE #454: Error in running PmfData::deallocate() on GPU
I was trying to use the `PmfData` class in my project. When running on GPU, an error occurred during `deallocate()`, saying that `CArena::free: unknown pointer !!!`.

Upon looking into the code, it seems that the error is caused by the fact that `m_data_d` is allocated with `The_Arena()` (line 48: `PMFData.H`) but is freed by `The_Device_Arena()` (line 61: `PMFData.H`). As a workaround, I substituted `The_Device_Arena()` with `The_Arena()`, so that the allocation and deallocation match. I have tested on both CPU and GPU, the issue appeared to be resolved.

I was wondering if this is a valid solution or if the line `The_Device_Arena()->free(m_data_d)` is intentional. I will be happy to contribute a pull request if the proposed solution aligns with the intended design.

Appreciate your insights. Thank you.

ensonun:
Closes #454

### DIFF
--- a/Utility/PMF/PMFData.H
+++ b/Utility/PMF/PMFData.H
@@ -58,7 +58,7 @@ public:
       amrex::The_Pinned_Arena()->free(m_data_h.pmf_Y);
     }
     if (m_device_allocated) {
-      amrex::The_Device_Arena()->free(m_data_d);
+      amrex::The_Arena()->free(m_data_d);
     }
   }
 


marchdf:
Thank you for opening this! When I think about it, I wonder if we should be allocating use `the_device_arena` and then deallocating in the same arena. But maybe if you've tested on the GPU/CPU combo then it doesn't matter...

marchdf:
We can ignore the failing CEPTR test, I fixed that in PR #452 .

baperry2:
@ensonun Thank you for submitting this fix!

@marchdf I think this is fine as is unless @jrood-nrel objects. 

We use `The_Arena` for prob_parm but `The_Device_Arena` for trans_parm. These should be an alias of each other anyway except when running with `amrex.the_arena_is_managed = 1` (https://amrex-codes.github.io/amrex/docs_html/GPU.html?highlight=the_arena_is_managed#memory-allocation).

baperry2 (approved):

jrood-nrel (approved):

jrood-nrel:
Thanks for this @ensonun . Maybe this is why the PMF files here never worked for us in PeleC?

ensonun:
Thanks all for approving this. It’s my pleasure being able to help!

marchdf:
@ensonun I think I created some conflicts with my latest commit (a bunch of file reorgs). Can you fix the conflict? Thank you!

PR #458, opened 2023-12-13
Fix warnings

marchdf:


baperry2 (approved):
Approving because this does not change the behavior of the code as written, but I do have some concern that there may be a bug in here

Source/Spray/BreakupSplash/WallFilm.H
```diff
@@ -158,20 +158,20 @@ calculateFilmSource(
...
+          // rho_film += Y_film[spf] / fdat.rhoL(T_film, spf);
         }
-        rho_film = 1. / rho_film;
-      } else {
-        rho_film = fdat.rhoL(T_film, 0);
-      }
+        // rho_film = 1. / rho_film;
+      } // else {
+        //  rho_film = fdat.rhoL(T_film, 0);
+      //}
```
marchdf, line 174: @ldowen do you know why these are computed but not used? Is that a bug or just something that changed?

ldowen: These were recomputed in case this whole update was in a sub-cycle. That clearly isn't the case so they are unnecessary.

PR #459, opened 2023-12-13
Move PeleRad source files to PelePhysics

jrood-nrel:
Tasks:

- [x] Remove std::filesystem
- [x] Add existing tests


marchdf:
when we merge can we add "Co-authored-by: Wenjun Ge <39968398+wjge@users.noreply.github.com>" to the commit message? That way we know who to bug for certain files ;)

baperry2:
@wjge pointed out that he set up some unit tests for radiation. Can we port those over as well?

wjge:
These six tests are needed: 

tstPOneMulti.cpp
tstPOneMultiAF.cpp
tstPOneMultiEB.cpp
tstPOneSingle.cpp
tstPOneSingleAF.cpp
tstPOneSingleEB.cpp

The boost test stuff can be replaced by what Pele is using.

jrood-nrel:
Yes. Currently I need to debug why moving these files causes SootRadTest to fail.

baperry2:
When I initially added that test to PeleLMeX, I had some mysterious failures that ended up being that it wasnt reading the kp data files correctly, so I'd recommend double checking that

Testing/Exec/Radiation/CMakeLists.txt
baperry2, line 1: Can we have these tests run in the CI?

jrood-nrel: In PeleLMeX we can. Otherwise it takes a lot of duplicate code to enable AMReX here. It could maybe happen in a reorganization of CMake.

baperry2: We should run them somewhere, otherwise they will break.

jrood-nrel: https://github.com/AMReX-Combustion/PeleLMeX/pull/326

marchdf (approved):

PR #462, opened 2023-12-19
Remove prob parm passing in spray

marchdf:


jrood-nrel (approved):

baperry2 (commented):
I think this should be ok.

PR #478, opened 2024-02-17
 surfchem - modified the ckinu method in the ceptr ck module

jAnirudh:
Writes an additional `SKINU` method in heterogeneous `mechanism.cpp` files. The `SKINU` method is analogous to `CKINU` and returns the species in a surface reaction, their indices and stoichiometric coefficients.

Files changed: `ck.py`, `converter.py`, non-qss mechanism files


baperry2 (approved):
These changes look fine to me.

@jAnirudh - as you add this capability, could you add a sample mechanism with surface chemistry to demonstrate the functionality?

jAnirudh:
Thanks for approving these changes @baperry2.

The issue with adding a surface chemistry mechanism at this stage is that ceptr crashes while generating the source code for heterogeneous mechanisms. This is because, apart from the few methods I've modified, the remaining methods don't know what to do when they come across surface species and reactions. Essentially, ceptr needs to be worked through completely to generate source codes for heterogeneous mechanisms.

I could, however, add the yaml file and the incomplete source code if that's helpful.

baperry2:
Oh ok, that makes sense not to include it yet then.

Issue #480, opened 2024-03-07
cvode maxsteps


ThomasHowarth:
Hi,
The maximum number of substeps for chemistry is hard-coded at 10000. I've found that for cases with stiff chemistry (particularly nitrogen chemistry) that this isn't sufficient, even in flows that are at a reasonable enough speed that (for LMeX) the timestep is on the order of 1e-7 < dt < 1e-6. Do you foresee issues in adding in an input to potentially increase this number? When running on GPUs the chemistry is usually cheaper than the diffusion solver anyway, so on initial testing it doesn't seem to add too much computational cost, and reduces the constant crashes I seem to be having. Happy to take other suggestions, but the timestep is sometimes required to be so small that I may as well run cases in PeleC, even though the flow is nowhere near the speed to necessitate that.

https://github.com/AMReX-Combustion/PelePhysics/blob/c2d6577ac92a303d5916ac827b191ea7e8b02ee5/Source/Reactions/ReactorCvode.cpp#L492

marchdf:
I like the idea. Feel free to put in a PR.

drummerdoc:
Hmm, nothing wrong with adding another parameter to the inputs, but when it takes that many inner steps to get there, I'll bet that there are better setting for SUNDIALS to solve that problem...e.g., if you are using the Newton-Krylov solver, maybe Newton with a Jacobian, a more robust linear solver and perhaps twiddled options for the preconditioned?

ThomasHowarth:
I'll try other solve_types and preconditioners to see what works best, but I've also PR'd this input. I'll also PR a small change to the LMeX documentation for this flag and include the suggestion of trying other solve_types before increasing the default. Thanks for the suggestions.

drummerdoc:
Cool, thanks Thomas!  If you can isolate one of the troublesome cells, might be useful to run it through ReactEval and play with the options until you dial something in.

PR #481, opened 2024-03-07
Max substep input for CVODE

ThomasHowarth:
A short PR to add the ability to specify number of substeps for the chemistry integration in CVODE. Addresses #480 . Will also add a small PR to LMeX documentation for the flag, along with a suggestion of trying other solve_types before increasing from the default.


baperry2 (approved):
Looks good.

PR #490, opened 2024-04-18
modify bounds checkings in the radiation module

wjge:
1. Fix an issue when the temperature is higher than 2800K;
2. Add a default value where there is no radiating gas.

### DIFF
--- a/Source/Radiation/SpectralModels.H
+++ b/Source/Radiation/SpectralModels.H
@@ -17,7 +17,7 @@ interpT(amrex::Real const& T, int& TindexL, amrex::Real& weight)
     return;
   }
   if (T > 2800) {
-    TindexL = 125;
+    TindexL = 124;
     weight = 1.0;
     return;
   }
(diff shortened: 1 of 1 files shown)


Source/Radiation/SpectralModels.H
```diff
@@ -58,6 +58,11 @@ getRadPropGas(
   amrex::GpuArray<amrex::Real, 126UL> const& kdatah2o,
   amrex::GpuArray<amrex::Real, 126UL> const& kdataco)
 {
+  if (
+    yco2(i, j, k) < 1.0e-5 && yh2o(i, j, k) < 1.0e-5 && yco(i, j, k) < 1.0e-5) {
+    absc(i, j, k) = 0.001;
+    return;
+  }
```
baperry2, line 65: Just want to make sure I get what's going on. Shouldn't this not be needed because later on this variable is limited?  https://github.com/AMReX-Combustion/PelePhysics/blob/80f2cfa0edd2e928a93bb9dff45e68d32d8699f7/Source/Radiation/PeleCRad.H#L125

wjge: The guard you showed failed in the EB production case (PeleLMeX case not PeleC) I've been testing. 
And the additional guard in this PR fixed the issue for some reason, guessing some EB cell might be the cause.

It might only need one guard, but now I didn't have a chance to test all my previous cases, will do that when possible.

baperry2: That's weird, but I guess it shouldn't hurt to have an extra guard. Although it might be a sign of some other unknown bug.

baperry2 (approved):

PR #491, opened 2024-04-19
Add cray to compilers for make TPL

baperry2:
AMReX gmake system supports cray compilers but we do not for building third party libraries, so this PR adds that to the ThirdParty GNUmakefile. Not sure if I missed any other changes would be needed. Seems to work on Kestrel.

### DIFF
--- a/ThirdParty/GNUmakefile
+++ b/ThirdParty/GNUmakefile
@@ -48,7 +48,12 @@ else
               ifeq ($(lowercase_hcomp),$(filter $(lowercase_hcomp),intel))
                 $(error for intel, specify COMP as intel-classic or intel-llvm)
               else
-                $(error Unknown COMP setting)
+                ifeq ($(lowercase_hcomp),$(filter $(lowercase_hcomp),cray))
+                   CCOMPILER = craycc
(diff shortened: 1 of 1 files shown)


ThirdParty/GNUmakefile
```diff
@@ -48,7 +48,12 @@ else
               ifeq ($(lowercase_hcomp),$(filter $(lowercase_hcomp),intel))
                 $(error for intel, specify COMP as intel-classic or intel-llvm)
               else
-                $(error Unknown COMP setting)
+                ifeq ($(lowercase_hcomp),$(filter $(lowercase_hcomp),cray))
```
marchdf, line 51: I have no problems with this but I gotta ask... Why do we have all these nested if statements? Make `else ifeq` is just the way things are done in gnu make.

marchdf (approved):

ThirdParty/GNUmakefile
baperry2: IDK haha. But given what's there just adding one more seems fine.

jrood-nrel (approved):

PR #492, opened 2024-04-24
Manifold Models: Tables and Network capability, wrappers in EOS, Transport

baperry2:
Includes new capability to define black box function (either a table or a neural network) and use this function in an EOS.

Co-authored by: @KiranEiden 

Does not support the Reactor code yet, that will be a separate PR.


Mechanisms/Null/mechanism.H
```diff
@@ -13,8 +13,13 @@
 /* Species */
 #define XO_ID 0
 
-#define NUM_ELEMENTS 1
+#ifdef USE_MANIFOLD_EOS
```
marchdf, line 16: Do you think it would be easier to just create a "manifold mech" to be used when using manifold stuff?

Source/Eos/EOS.H
```diff
@@ -20,6 +24,8 @@ using EosType = eos::GammaLaw;
 using EosType = eos::Fuego;
 #elif USE_SRK_EOS
 using EosType = eos::SRK;
+#elif USE_MANIFOLD_EOS && !defined(AMREX_USE_SYCL)
```
marchdf, line 27: are there other abort elsewhere that disallow sycl for manifold? like in the gnu make or cmake paths?

Source/Eos/EosParams.H
```diff
@@ -0,0 +1,176 @@
+#ifndef EOS_PARAMS_H
+#define EOS_PARAMS_H
+
+#include <AMReX_REAL.H>
+#include <AMReX_ParmParse.H>
+
+#include "mechanism.H"
+#include "BlackBoxFunc.H"
```
marchdf: Instead of `Func` can we be verbose and say `Functions` ?

Source/Eos/EosParams.H
```diff
@@ -0,0 +1,176 @@
...
+
+#ifndef MANIFOLD_DIM
+#define MANIFOLD_DIM 1
+#endif
+
+#ifndef AMREX_USE_SYCL
+template <>
+struct EosParm<Manifold>
+{
+  const BBFuncData* manf_data;
```
marchdf: In some other places, sometimes, inconsistently we prefix class member vars with `m_`. Could do that here.

Source/Eos/EosParams.H
```diff
@@ -0,0 +1,176 @@
...
+  amrex::Real Pnom_cgs;
+  density_lookup_type dens_lookup;
+  bool compute_temperature{false};
+  int idx_density{0};
+  int idx_T{0};
+  int idx_Wdot[MANIFOLD_DIM];
+  bool has_mani_src = true;
```
marchdf: change to `has_manifold_src{true}`

Source/Eos/EosParams.H
```diff
@@ -0,0 +1,176 @@
...
+
+    // For manifold table parameter source terms, assume if index not found,
+    // source term is 0 For neural net, require a definition to be supplied for
+    // each manifold parameter
+    if (parm_in->m_h_parm.has_mani_src) {
+      for (int idim = 0; idim < h_manf_data_in->Ndim; idim++) {
+        std::string dimname = std::string(
+          &h_manf_data_in->dimnames[idim * h_manf_data_in->len_str],
+          h_manf_data_in->len_str);
+        std::string dim_src;
```
marchdf: could be const std::string with ternary expression

Source/Eos/Fuego.H
```diff
@@ -16,6 +16,38 @@ struct Fuego
 
   static std::string identifier() { return "Fuego"; }
 
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  static void RY2R(const amrex::Real RY[], amrex::Real& R)
+  {
+    R = 0.0;
+    for (int is = 0; is < NUM_SPECIES; is++) {
```
marchdf: using `is` is a slighly different style than `i` for for loops on species in this file. just noting this in case you want to do anything about it. I don't mind either way

Testing/Exec/TranEval/GPU_misc.H
```diff
@@ -37,11 +37,20 @@ initialize_data(
     L[n] = phi[n] - plo[n];
     P[n] = L[n] / 4.0;
   }
+#ifdef USE_MANIFOLD_EOS
```
marchdf, line 40: Are these ifdefs going to be all over our app codes do you think?

Testing/Exec/Make.PelePhysics
```diff
@@ -54,6 +54,20 @@ endif
 ifeq ($(Eos_Model),$(filter $(Eos_Model),Soave-Redlich-Kwong))
    DEFINES += -DUSE_SRK_EOS
 endif
```
marchdf, line 56: Like in here some kind of sycl manifold check

Support/CMLM/peletable_trans_2dim_10grid
marchdf, line 1: why is this showing as a binary file? are the permissions +x on them? same for the one above.

Mechanisms/Null/mechanism.H
baperry2: I did that initially but I think this makes more sense. This is the only change required, so it avoids a lot of code duplication. Because the Null mechanism doesn't get updated automatically when changes are made in ceptr, this reduces the number of manual changes that must be made.

Support/CMLM/peletable_trans_2dim_10grid
baperry2: It's meant to be a binary, but not executable. It uses the same format as the chemtable files from NGA, btw. Binary because the files become quite large for higher table dimensions. Eventually I'd like to remove these and have creating the files become part of the testing workflow, but that will come later. For now I just included some small files to be able to run tests. They're here because they're shared between multiple tests and I didn't want redundant copies.

Testing/Exec/TranEval/GPU_misc.H
baperry2: None should be required for PeleC because it won't use this capability. They will pop up in a few places in LMeX, but mostly just case data initialization similar to here. For the most part, adding an EOS function that computed Rho from the RhoYs removes the need for ifdefs.

Source/Eos/EosParams.H
baperry2: I don't think that is necessary here. This is a simple struct that only exists as a data container with no member functions, so there shouldn't be any ambiguity (all data are members). Similar to transparm, where we also don't prefix with `m_`

Source/Eos/EOS.H
baperry2: This should catch the condition regardless of build system and PelePhysics/C/LMeX. I think it would take too many checks to do it in both build systems for all application codes.

marchdf (approved):

PR #499, opened 2024-05-15
Update the CEPTR usage doc slightly.

EnnaDelfen:
Hi Marc, here's the PR with a couple modifs to the CEPTR documentation. 
I reorganized the section a little for it to follow the actual steps performed (generate a yaml first, then generate the single mechanisms file, then maybe to the batch generation). 
I also made sure the arguments given in the code snippets are the same as in the text because it was confusing me at times. 
I've tried clarifying the commands for generating a single QSS YAML file and then I wanted to explain the steps for generating a single QSS Pele mechanism file BEFORE the batch generation is mentioned - but then I found that there's an entire other section dedicated to that so I let the reference to it and simply moved the QSS section before the batched section.
Finally I warned users that one may need a certain Python version/Cantera to be installed.
Overall, very little but I think it's a tad improved.

### DIFF
--- a/Docs/sphinx/Ceptr.rst
+++ b/Docs/sphinx/Ceptr.rst
@@ -110,7 +110,7 @@ The full list of options is::
 
 For a detailed description of these options and a further information on the way QSS mechanism are treated in CEPTR the reader may consult :ref:`the QSS section <sec_qss>`.
 
-To generate a QSS C++ mechanism from the ``.yaml`` file thus created, taylored to your needs, please refer to Tutorials :ref:`Generating NC12H26 QSS mechanism with analytical jacobian <sec_tutqss1>` and :ref:`Generating NC12H26 QSS mechanism without analytical jacobian <sec_tutqss2>`.
+To generate a QSS C++ mechanism from the ``.yaml`` file thus created, tailored to your needs, please refer to Tutorials :ref:`Generating NC12H26 QSS mechanism with analytical jacobian <sec_tutqss1>` and :ref:`Generating NC12H26 QSS mechanism without analytical jacobian <sec_tutqss2>`.
 
 Batched generation of Pele-compatible mechanisms 
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
--- a/Docs/sphinx/Ceptr.rst
+++ b/Docs/sphinx/Ceptr.rst
@@ -19,7 +19,7 @@ To install CEPTR dependencies::
   $ cd ${PELE_PHYSICS_HOME}/Support/ceptr
   $ poetry update
 
-.. note:: Note that you will need to use a Python version >=3.10 and <3.12. If a compatible version exists in your system then poetry will try to find and use it. Otherwise, think about using a `conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`_ environment to manage packages and their dependencies without tampering with your system. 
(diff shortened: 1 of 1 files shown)


Docs/sphinx/Ceptr.rst
```diff
@@ -53,13 +67,53 @@ There are three ways to use CEPTR to generate C++ mechanism files for a given ch
...
+   .. note:: It is possible that using this option will require for you to have a valid Cantera installed somewhere. Again, we strongly suggest using a `conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`_ environment to install all required package. A simple ceptr environment can be generated using the following yaml script:: 
+
+     $ name: ceptr
+     $ channels:  
+     $ - conda-forge 
+     $ dependencies: 
+     $ - python=3.10  
+     $ - cantera
+     $ - cython  
+     $ - numpy
```
malihass: This seems to repeat cantera and numpy dependencies compared with  `pyproject.toml`
I'd prefer if we keep the conda install dependencies to a minimum:

Could you please try to see if creating you conda env with `conda create --name ceptr python=3.10 poetry` before calling `convert.sh` does the trick?

malihass (changes requested):

Docs/sphinx/Ceptr.rst
```diff
@@ -19,13 +19,27 @@ To install CEPTR dependencies::
   $ cd ${PELE_PHYSICS_HOME}/Support/ceptr
   $ poetry update
 
+.. note:: Note that you will need to use a Python version >=3.10 and <3.12. If a compatible version exists in your system then poetry will try to find and use it. Otherwise, think about using a `conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`_ environment to manage packages and their dependencies without tampering with your system.
```
baperry2: How about saying to check Support/ceptr/pyproject.toml for the latest supported python versions in case they get updated and we forget to modify this text in the future?

Docs/sphinx/Ceptr.rst
EnnaDelfen: As always, I can't seem to reproduce the errors that I had now that I've tried/installed all these various environments. I know for a fact that simply invoking the convert.sh from the Mechanism directory directly (LiDryer) gave me , first a python and then a cantera error. Which I solved by including cantera in the ceptr env; and which was the entire point of my modifying the instructions actually  ... if we remove that then, yes I've slightly improved the doc but I'm not addressing the main issue I encountered which is a lack of Cantera that somehow wasn't automatically taken care of.

malihass: Clearly the solution you provided is working, but it is not ideal since `pyproject.toml` is supposed to already take care of the cantera dependency.
I just tried what I suggested on kestrel on a fresh conda  env and it seems to work, but it'd be good if we can confirm that it does take care of this issue. 
@drummerdoc you mentioned you had the same problem as @EnnaDelfen , could you maybe please try what I mentioned?

Docs/sphinx/Ceptr.rst
EnnaDelfen: Done.

Docs/sphinx/Ceptr.rst
EnnaDelfen: I've tried everything on another (older) mac system and ran into issues obtaining cantera via the poetry channel. This time the issue was a bit different in that the poetry update clearly indicated that cantera was not found and thus not installed. So the `conda create --name ceptr python=3.10 poetry ` alone doesn't fix the issue. Again the only fix I could find was to use the conda-forge channel to create a conda env that includes cantera. I've explained everything to Malik in private. The source of the issue may not be completely identified but the proposed note (which I have simplified to python 3.10 and cantera, no need for numpy or cython) did fix all my issues so far. But you can go ahead and ignore it if you prefer it.

malihass: Thanks for checking all that!

malihass (approved):

baperry2 (approved):

PR #504, opened 2024-05-27
Chemical mask implementation

SreejithNREL:
Chemistry mask implementation.

Updated Arkode, Cvode, RK64, and BDF chem integrators to mask chemical reactions inside a user-specified box region. Verified code change on PMF case for Clang serial, MPI, CUDA, and HIP.

### DIFF
--- a/Source/Reactions/ReactorArkode.H
+++ b/Source/Reactions/ReactorArkode.H
@@ -129,23 +129,23 @@ public:
   }
 
   void unflatten(
-      const amrex::Box& box,
-      const int ncells,
-      amrex::Array4<amrex::Real> const& rhoY,
-      amrex::Array4<amrex::Real> const& temperature,
-      amrex::Array4<amrex::Real> const& rhoE,
-      amrex::Array4<amrex::Real> const& frcEExt,
-      amrex::Array4<amrex::Real> const& FC_in,
-	  amrex::Array4<int> const& mask,
-      amrex::Real* y_vect,
-      amrex::Real* vect_energy,
-      long int* FCunt,
-      amrex::Real dt)
-    {
-      flatten_ops.unflatten(
-        box, ncells, m_reactor_type, m_clean_init_massfrac, rhoY, temperature,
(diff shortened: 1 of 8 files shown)


marchdf:
Let's ignore codespell for now. We need to fix that in a different PR.

hsitaram (commented):
I am a bit worried though, shouldn't the states be updated with the adv/diffusion sources * dt before being sent back to the application codes (LM/C)?

Source/Reactions/ReactorArkode.H
```diff
@@ -128,6 +128,25 @@ public:
...
 
+  void unflatten(
+    const amrex::Box& box,
+    const int ncells,
+    amrex::Array4<amrex::Real> const& rhoY,
+    amrex::Array4<amrex::Real> const& temperature,
+    amrex::Array4<amrex::Real> const& rhoE,
+    amrex::Array4<amrex::Real> const& frcEExt,
+    amrex::Array4<amrex::Real> const& FC_in,
+    amrex::Array4<int> const& mask,
```
marchdf, line 139: the mask should probably come in as `amrex::Array4<const int> const&`

Source/Reactions/ReactorArkode.cpp
```diff
@@ -135,7 +135,7 @@ ReactorArkode::react(
   amrex::Array4<amrex::Real> const& rEner_in,
   amrex::Array4<amrex::Real> const& rEner_src_in,
   amrex::Array4<amrex::Real> const& FC_in,
-  amrex::Array4<int> const& /*mask*/,
+  amrex::Array4<int> const& mask,
```
marchdf, line 138: same here.

Source/Reactions/ReactorRK64.cpp
```diff
@@ -246,72 +246,77 @@ ReactorRK64::react(
       rYsrc_ext[sp] = rYsrc_in(i, j, k, sp);
     }
 
-    int nsteps = 0;
```
marchdf, line 249: did anything happen here beyond just a big indent because of the if condition?

marchdf (changes requested):

PR #507, opened 2024-06-06
peleparams: common data container for PMF, transport (others in future)

baperry2:
Takes a piece from #492 to make reviewing easier.

PeleParams is a class template that allows copying arbitrary data structures from host to device, taking advantage of The_Pinned_Arena (simple data structures of fixed size can be copied directly without dealing with that though). An InitParm class must be defined with a function that populates the data structure (and cleans it up, if needed). There's some funky stuff to be able to use PeleParams to create derived versions of a base data structure, but for most use cases that can be transparently ignored. Written as a generalization of the old PMFData structure.

Will require minor downstream changes in LMeX/C.

Also implements bugfix in PMF interpolation from https://github.com/AMReX-Combustion/PeleC/pull/808

### DIFF
--- a/Source/Transport/TransportParams.H
+++ b/Source/Transport/TransportParams.H
@@ -4,9 +4,11 @@
 #include <AMReX_REAL.H>
 #include <AMReX_ParmParse.H>
 #include "TransportTypes.H"
+#include "PeleParamsGeneric.H"
 #include "EOS.H"
 
-namespace pele::physics::transport {
+namespace pele::physics {
+namespace transport {
 
 struct ConstTransport;
 struct SimpleTransport;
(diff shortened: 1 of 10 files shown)


Source/Transport/TransportParams.H
```diff
@@ -4,9 +4,11 @@
 #include <AMReX_REAL.H>
 #include <AMReX_ParmParse.H>
 #include "TransportTypes.H"
+#include "PeleParamsGeneric.H"
 #include "EOS.H"
 
-namespace pele::physics::transport {
+namespace pele::physics {
+namespace transport {
```
marchdf, line 11: I think this is a clang tidy warning? Someone out there is telling us to `pele::physics::transport` instead of breaking it out? Or maybe I am missing something

marchdf (commented):
Nicely done!!

Source/Transport/TransportParams.H
baperry2: clang-tidy does require that nested namespaces be grouped when possible. However, it's not possible in this case so it's not an error. The transport namespace doesn't encompass everything in the file now. The InitParm template specializations have to be in the same namespace as the initial template definition, and that has to be the pele::physics namespace because the same template is used across different modules of PelePhysics.

marchdf (approved):

PR #513, opened 2024-07-19
Support Manifold EoS in RK64 Reactor

baperry2:
- Pass eosparm through the ReactorRK64 call stack so it can be used as needed
- Also fixes a bug where the limit to avoid overstepping the integration time for RK64 was not being applied properly for the 1D version.
- add a test case for ignition delay with tabulated chemistry


Source/Reactions/ReactorBase.H
```diff
@@ -81,11 +81,28 @@ public:
...
   void set_typ_vals_ode(const std::vector<amrex::Real>& ExtTypVals);
 
+  // Manifold EOS needs an eosparm - right now that is only propagated through
+  // the RK64 reactor
+  virtual void set_eos_parm(
+    const pele::physics::eos::EosParm<pele::physics::PhysicsType::eos_type>*
+      eosparm)
+  {
+    auto eos = pele::physics::PhysicsType::eos(eosparm);
+    if (eos.identifier() == "Manifold") {
```
marchdf, line 91: could replace all this with `AMREX_ALWAYS_ASSERT_WITH_MESSAGE`

marchdf (approved):

marchdf:
In this PR, can you fix a small typo? https://github.com/AMReX-Combustion/PelePhysics/blob/development/README.md?plain=1#L18 should be Radiative

Testing/Exec/IgnitionDelay/check_ignDelay.py
```diff
@@ -18,7 +18,7 @@ def readResult(filename):
     tIgn,uncertainty = readResult('log')
 
     # Compare to theoretical value
-    if abs(tIgn-0.0763)>10*uncertainty:
+    if abs(tIgn-0.0763)>15*uncertainty:
```
malihass: I think you are doing this because you want the test to pass with the manifold.

Could you please maybe change check_ignDelay to receive 2 arguments (with argparse): 1 is the expected ignition delay (0.0763) and 1 would the tolerance? Then you can change the tolerance for the manifold test.

Testing/Exec/IgnitionDelay/computeIgnitionDelay.py
```diff
@@ -72,33 +72,33 @@ def readMultiColFile(filename,headerSize=0):
...
         writeNewInput(args.inputFile,'inputs/inputs.0d_refine',ignitionDelayPP*1.1,ndt*2)
-    
-    if args.refine: 
-        logMessage = f"T = {temp:.3g} K , P = {press/101325.0:.3g} atm, Phi = {phi:.3g},  tIgn = {ignitionDelayPP:.3g} +/- {dt/ndt:.3g} s"
+
+    if args.refine:
+        logMessage = f"T = {temp:.3g} K , P = {press/1013250.0:.3g} atm, Phi = {phi:.3g},  tIgn = {ignitionDelayPP:.3g} +/- {dt/ndt:.3g} s"
         print(logMessage)
-        writeResult('log',temp,press,phi,ignitionDelayPP,dt/ndt) 
+        writeResult('log',temp,press,phi,ignitionDelayPP,dt/ndt)
```
malihass, line 102: Could you please check if the 101325 factor is also messed up in `writeResult`?

malihass (changes requested):
Only minor comments, the manifold seems to represent ignition delay really well!

Testing/Exec/IgnitionDelay/check_ignDelay.py
baperry2: Done.

Testing/Exec/IgnitionDelay/computeIgnitionDelay.py
baperry2: Could call, it also needed correction there.

malihass (approved):

PR #515, opened 2024-07-22
Simplify Make.package in mechanisms

baperry2:
Get rid of a bunch of extraneous/deprecated stuff in the Make.package files for each mechanism.

### DIFF
--- a/Mechanisms/Aromatic_KrNara/Make.package
+++ b/Mechanisms/Aromatic_KrNara/Make.package
@@ -1,9 +1,1 @@
-CEXE_headers+=mechanism.H
-
 CEXE_sources+=mechanism.cpp
-
-LIBRARIES +=
-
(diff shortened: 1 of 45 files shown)


marchdf:
Ah wait there's a reason for some of these. I remember trying to remove them and then @esclapez said something that made me change my mind. Any chance you can repeat that info here @esclapez ?

baperry2:
I'm pretty sure the PMFs directory is hold over from a long long time ago when PMF data was compiled as source code. The rest of it I think should be ok as long as everything compiles, although I didn't verify that for LMeX/C, will do that now. But @esclapez feel free to jump in if there's something I'm missing.

marchdf (approved):

baperry2:
Going to merge this because it doesn't appear to cause any issues building C or LMeX.

PR #518, opened 2024-08-05
Update amrex

marchdf:


baperry2 (approved):
While doing this, let's also roll back #510 to address the issue described in  https://github.com/AMReX-Combustion/PeleC/issues/830 and get dependabot updates working again. (see here for log of dependabot failure, I believe, but have not confirmed, that this is occurring because of the shwllow submodule clones: https://github.com/AMReX-Combustion/PelePhysics/network/updates/863923305).

Due to a bug in git (https://github.com/git/git/pull/1740), the default shallow clones are not saving us that much download size anyway.

PR #536, opened 2024-10-09
Make EOS functions templates to allow for CellData as an argument

baperry2:
The idea here is to enable simplifying EOS calls. We have a lot of cases where we create an array of mass fractions from Array4 data just to pass it to an EOS function. Instead, we can modify the EOS functions so they can take `amrex::CellData` as an argument (really, any data structure that can be indexed with the `[]` operator and return an `amrex::Real &`). As an example,

```
amrex::Real rhospec[NUM_SPECIES] = {0.0};
amrex::Real rho = 0.0;
for (int n = 0; n < NUM_SPECIES; ++n) {
    rhospec[n] = rhoY(i,j,k,n);
}
eos.RY2R(rhospec, rho);
```
becomes

```
amrex::Real rho = 0.0;
eos.RY2R(rhoY.cellData(i,j,k), rho);
```

The `static_asserts` aren't strictly necessary but serve a readability purpose to show that `T` in the templates must be something that looks like an array of `amrex::Real`s.

### DIFF
--- a/Source/Eos/Fuego.H
+++ b/Source/Eos/Fuego.H
@@ -16,27 +16,27 @@ struct Fuego
 
   static std::string identifier() { return "Fuego"; }
 
-  template <typename T, int first = 0>
+  // ReallArrayLike can be anything with an [] operator that returns
+  // amrex::Real&
+  template <typename RealArrayLike>
   AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE static void
-  RY2R(const T RY, amrex::Real& R)
+  RY2R(const RealArrayLike& RY, amrex::Real& R, int first = 0)
   {
-    static_assert(
-      std::is_same_v<decltype(RY[0]), amrex::Real&> ||
-      std::is_same_v<decltype(RY[0]), const amrex::Real&>);
     R = 0.0;
     for (int i = 0; i < NUM_SPECIES; i++) {
       R += RY[first + i];
     }
   }
 
-  template <typename T, typename T2>
-  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE static void
-  RY2RRinvY(const T RY, amrex::Real& R, amrex::Real& Rinv, T2 Y)
+  // ReallArrayLike can be anything with an [] operator that returns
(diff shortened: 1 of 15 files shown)


Source/Eos/Fuego.H
```diff
@@ -16,21 +16,27 @@ struct Fuego
 
   static std::string identifier() { return "Fuego"; }
 
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  static void RY2R(const amrex::Real RY[], amrex::Real& R)
+  template <typename T, int first = 0>
+  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE static void
+  RY2R(const T RY, amrex::Real& R)
```
marchdf: does this need to be `const T& RY` to avoid a copy?

marchdf:
I love this idea!

Source/Eos/Fuego.H
baperry2: Yes, good catch.

baperry2:
So making this apply across all the EOS functions would require also doing the same templating in all the functions that get generated by CEPTR. I don't think it's a priority to do that right now, so I'm only going to do a couple of the EOS functions like `RY2R` that are the ones I actually need this for and don't use any functions generated by CEPTR.

Source/Eos/Fuego.H
```diff
@@ -16,20 +16,26 @@ struct Fuego
...
   static std::string identifier() { return "Fuego"; }
 
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  static void RY2R(const amrex::Real RY[], amrex::Real& R)
+  // ReallArrayLike can be anything with an [] operator that returns
+  // amrex::Real&
+  template <typename RealArrayLike>
+  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE static void
+  RY2R(const RealArrayLike& RY, amrex::Real& R, int first = 0)
```
marchdf, line 23: you can const first in all of these I think.

marchdf (approved):

PR #539, opened 2024-10-31
Unit conversions

d-montgomery:
This pull request introduces a new namespace called `utilities`, which contains a struct for unit conversion functions. This will primarily benefit users of PeleLMeX when calling EOS functions from PelePhysics, as the units in PeleLMeX are in MKS.  The `unit` functions are return type `amrex::Real` for ease of passing into the EOS functions from PeleLMeX.  

You can test it out on a local case in PeleLMeX  by adding the following to the `Make.PeleLMeX` file:
```
PPdirs += Source/Utility/Utilities
```
The functions are called similar to the eos functions.  For example:
```
auto eos = pele::physics::PhysicsType::eos();
auto units = pele::physics::utilities::units();
eos.PYT2R(units.MKS2CGS_P(p), massfrac, T, rho_cgs);
rho = units.CGS2MKS_RHO(rho_cgs);
```

I look forward to your feedback.

### DIFF
--- a/Source/Utility/Utilities/UnitConversions.H
+++ b/Source/Utility/Utilities/UnitConversions.H
@@ -6,238 +6,250 @@
 #include <AMReX_Utility.H>
 
 namespace pele::physics::utilities {
-
+namespace cgs2mks {
 // CGS to MKS conversions
-struct cgs2mks
+template <int n = 0>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real
+Length(const amrex::Real L_cgs)
 {
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real Length(const amrex::Real L_cgs)
-  {
+  // Usage: Length(L_cgs) or Length<n>(L_cgs)
+  if constexpr (n == 0) {
     return L_cgs * 1.0e-2; // cm to m
-  }
-
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real Length_Pow(const amrex::Real L_cgs, const amrex::Real n)
-  {
+  } else {
     return L_cgs * std::pow(1.0e-2, n); // cm^n to m^n
   }
+}
 
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real Mass(const amrex::Real M_cgs)
-  {
+template <int n = 0>
+AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real
+Mass(const amrex::Real M_cgs)
+{
+  // Usage: Mass(M_cgs) or Mass<n>(M_cgs)
+  if constexpr (n == 0) {
     return M_cgs * 1.0e-3; // g to kg
-  }
-
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real Mass_Pow(const amrex::Real M_cgs, const amrex::Real n)
-  {
+  } else {
     return M_cgs * std::pow(1.0e-3, n); // g^n to kg^n
   }
+}
 
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real Energy(const amrex::Real E_cgs)
-  {
-    return E_cgs * 1.0e-7; // erg to J
-  }
-
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real Rho(const amrex::Real rho_cgs)
-  {
-    return rho_cgs * 1.0e3; // g/cm^3 to kg/m^3
-  }
-
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real U(const amrex::Real u_cgs)
-  {
-    return u_cgs * 1.0e-2; // cm/s to m/s
-  }
-
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real RhoU(const amrex::Real rhou_cgs)
-  {
-    return rhou_cgs * 1.0e1; // g/cm^2/s to kg/m^2/s
-  }
-
-  AMREX_GPU_HOST_DEVICE
-  AMREX_FORCE_INLINE
-  amrex::Real P(const amrex::Real p_cgs)
-  {
-    return p_cgs * 1.0e-1; // dyne/cm^2 to Pa
-  }
-
-  AMREX_GPU_HOST_DEVICE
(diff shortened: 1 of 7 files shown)


Source/Utility/Utilities/UnitConversions.H
```diff
@@ -0,0 +1,241 @@
...
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  amrex::Real CGS2MKS_LENGTH(const amrex::Real L_cgs)
+  {
+    return L_cgs * 1.0e-2; // cm to m
+  }
+
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  amrex::Real CGS2MKS_LENGTH_POW(const amrex::Real L_cgs, amrex::Real n)
```
marchdf: const the second arg? Same for a bunch of others in this file.

marchdf:
I like where this is going! Would it be worth to have 2 structs? `mks2cgs` and `cgs2mks`, each encapsulating their own functions? That way the call signature can be shorter and maybe clearer? 
```
auto eos = pele::physics::PhysicsType::eos();
auto m2c = pele::physics::utilities::mks2cgs();
eos.PYT2R(m2c.P(p), massfrac, T, rho_cgs);
auto c2m = pele::physics::utilities::cgs2mks();
rho = c2m.RHO(rho_cgs);
```

Source/Utility/Utilities/UnitConversions.H
d-montgomery: Ah yes, good catch.

d-montgomery:
> I like where this is going! Would it be worth to have 2 structs? `mks2cgs` and `cgs2mks`, each encapsulating their own functions? That way the call signature can be shorter and maybe clearer?

I thought about this as well, but never wrote it out.  Now that I see it, I agree that two structs is a lot cleaner.  I'll push through an updated version in a bit.  Thanks for the suggestion.

marchdf:
Though I am curious, any reason to have a struct at all? Just plain functions might be fine? You could encapsulate in a namespaces:
```
eos.PYT2R(pele::physics::utilities::mks2cgs::P(p), massfrac, T, rho_cgs);
```
and if the verbosity bothered you:
```
// somewhere in a header
using m2c = pele::physics::utilities::mks2cgs

// actual call in code
eos.PYT2R(m2c::P(p), massfrac, T, rho_cgs);
```

d-montgomery:
> Though I am curious, any reason to have a struct at all? Just plain functions might be fine? You could encapsulate in a namespaces:

I also thought about this, but wanted to match the behavior of the eos calls.  I'm happy either way as long as I don't have to think about unit conversions!

marchdf:
I guess the advantage of plain functions is that you aren't allocating an object with a bunch of functions that you aren't going to use. You also remove a line of code because you don't need a "unit" variable. I did the `eos` thing because of templating and sometimes member variables. I don't see that being necessary here? @bperry65 correct me if I am not thinking this through properly.

baperry2:
I agree with @marchdf on using a namespace rather than a struct. Simplifies the code and might be a tad more efficient. 

It would be more or less equivalent to make all the function static struct members, which clang-tidy will insist upon anyway and lets them be called without instantiating an object of the struct, so that's another option as well.

Source/Utility/Utilities/UnitConversions.H
```diff
@@ -0,0 +1,243 @@
...
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  amrex::Real Length(const amrex::Real L_cgs)
+  {
+    return L_cgs * 1.0e-2; // cm to m
+  }
+
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  amrex::Real Length_Pow(const amrex::Real L_cgs, const amrex::Real n)
```
baperry2: maybe this is just my recent obsession with templating things, but I think this could be more efficient by having `n` be a template parameter rather than an argument, as then you'd be able to compute the `pow` in a `constexpr` at compile time rather than on every call at runtime.

marchdf: I love that ;)

marchdf:
static member functions works for me as well. The only advantage with the namespace would be maybe a little less lines of code.

Source/Utility/Utilities/UnitConversions.H
d-montgomery: > I think this could be more efficient by having `n` be a template parameter rather than an argument, as then you'd be able to compute the `pow` in a `constexpr` at compile time rather than on every call at runtime.

I got rid of the Length_Pow function and edited the Length function:
```
template <int n = 0>
AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real
Length(const amrex::Real L_cgs)
{
  // Usage: Length(L_cgs) or Length<n>(L_cgs)
  if constexpr (n == 0) {
    return L_cgs * 1.0e-2; // cm to m
  } else {
    return L_cgs * std::pow(1.0e-2, n); // cm^n to m^n
  }
}
```

Also, I got rid of the `struct` and created namespaces as @marchdf suggested.

Source/Utility/Utilities/Utilities.H
baperry2, line 1: Do we need this file?

Source/Utility/Utilities/Utilities.cpp
baperry2, line 1: Do we need this file?

d-montgomery: Technically no, but it is useful if other utilities are added later.

baperry2: Ok. Let's keep `Utilities.H` for that reason so that the includes for the headers are set up, but get rid of `Utilities.cpp` if that works

Source/Utility/Utilities/UnitConversions.H
baperry2: Hmm I dug into this a bit and it looks like what I had in mind is not possible because `std::pow` is not constexpr capable. Also prior to C++20 you can't template on floats so this doesn't allow fractional exponents. I'd say go back to your original implementation, but instead of calling the power version `Length_Pow`, just have it overload `Length`. Sorry for misleading you.

d-montgomery: No worries.  I didn't know you couldn't template on floats until I tried, so this was a good exercise at least.  I'll push the updated version through soon.

baperry2 (approved):

drummerdoc:
Were we able to completely side-step the pre-exponential factor crap in the chemical production rates associated with units?  That was always icky...maybe it was fixed long ago....

baperry2:
I'm sure there's a bunch of ickiness with pre-exponential factor units stuff within CEPTR. But because the interfaces to the flow solvers only happens through the species net production rates (and even the Jacobian is in those terms), we don't have to deal with that anywhere for cgs -> mks conversions in PeleLMeX.

PR #542, opened 2024-11-05
speciesNames for Manifold EOS

baperry2:
These get read from the table metadata rather than specified at compile time. Therefore, to define them, eosparm needs to be passed to the appropriate function. The implementation should be backward compatible and only require the modified call signature when using Manifold EOS.

This PR also modifies the out-of-class EOS function templates work to eliminate `EOS.cpp` and reduce redundant code, but perhaps is less readable? Open to suggestion on this.


marchdf:
I am fine with this moving the definitions to the header. The only thing is making sure the linkers don't hit "multiple definition" type of errors.

baperry2:
The multiple definition error is resolved by marking the template specializations as `inline`

marchdf (approved):

PR #543, opened 2024-11-05
Flow Probes

SreejithNREL:
Created a new diagnostic to probe flow variables (user-defined) at user-specified step intervals. The ASCII output data is written to "./temporals/" folder.

### DIFF
--- a/Source/Utility/Diagnostics/DiagProbe.cpp
+++ b/Source/Utility/Diagnostics/DiagProbe.cpp
@@ -19,15 +19,18 @@ LinearInterpolate(
   amrex::Real slp[3] = {0.0};
 
   for (int cnt = 0; cnt < AMREX_SPACEDIM; cnt++) {
-      slp[cnt] = (xp[cnt] - x_low[cnt]) / dx[cnt];
-    }
+    slp[cnt] = (xp[cnt] - x_low[cnt]) / dx[cnt];
+  }
 
-  value += (1.0 - slp[0]) * (1 - slp[1]) * (1 - slp[2]) * neighbour_cells(0, 0, 0);
+  value +=
+    (1.0 - slp[0]) * (1 - slp[1]) * (1 - slp[2]) * neighbour_cells(0, 0, 0);
   value += slp[0] * (1 - slp[1]) * (1 - slp[2]) * neighbour_cells(0 + 1, 0, 0);
-  value += (1.0 - slp[0]) * slp[1] * (1 - slp[2]) * neighbour_cells(0, 0 + 1, 0);
+  value +=
+    (1.0 - slp[0]) * slp[1] * (1 - slp[2]) * neighbour_cells(0, 0 + 1, 0);
   value += slp[0] * slp[1] * (1 - slp[2]) * neighbour_cells(0 + 1, 0 + 1, 0);
 
-  value += (1.0 - slp[0]) * (1 - slp[1]) * slp[2] * neighbour_cells(0, 0, 0 + 1);
+  value +=
+    (1.0 - slp[0]) * (1 - slp[1]) * slp[2] * neighbour_cells(0, 0, 0 + 1);
   value += slp[0] * (1 - slp[1]) * slp[2] * neighbour_cells(0 + 1, 0, 0 + 1);
   value += (1.0 - slp[0]) * slp[1] * slp[2] * neighbour_cells(0, 0 + 1, 0 + 1);
   value += slp[0] * slp[1] * slp[2] * neighbour_cells(0 + 1, 0 + 1, 0 + 1);
@@ -258,7 +261,8 @@ DiagProbe::processDiag(
 #if (AMREX_SPACEDIM == 1)
           {
             neighbour_cells(0, 0, 0) = state(low_cell_idx_d[0], 0, 0, stIdx);
-            neighbour_cells(1, 0, 0) = state(low_cell_idx_d[0] + 1, 0, 0, stIdx);
+            neighbour_cells(1, 0, 0) =
+              state(low_cell_idx_d[0] + 1, 0, 0, stIdx);
           }
 #elif (AMREX_SPACEDIM == 2)
 	        {
@@ -281,7 +285,8 @@ DiagProbe::processDiag(
 	        }
 #endif
           tmp_values_d[n] = LinearInterpolate(
-            m_probe_loc_d, x_low_cell_d, neighbour_cells, dx_finest_lev_probe_d);
+            m_probe_loc_d, x_low_cell_d, neighbour_cells,
+            dx_finest_lev_probe_d);
         } else if (m_interpType_d == CellCenter) {
 #if (AMREX_SPACEDIM == 1)
           tmp_values_d[n] = state(m_probe_idx_d[0], 0, 0, stIdx);
--- a/Source/Utility/Diagnostics/DiagProbe.cpp
+++ b/Source/Utility/Diagnostics/DiagProbe.cpp
@@ -11,37 +11,26 @@ amrex::Real
 LinearInterpolate(
   amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> xp,
   amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> x_low,
-  amrex::Array3D<amrex::Real, 0, 1, 0, 1, 0, 1> cell_data,
+  amrex::Array3D<amrex::Real, 0, 1, 0, 1, 0, 1> neighbour_cells,
   amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx)
 {
-  amrex::Real alpha, beta, gama;
+
   amrex::Real value = 0.0;
+  amrex::Real slp[3] = {0.0};
 
-  alpha = 0.0;
-  beta = 0.0;
-  gama = 0.0;
+  for (int cnt = 0; cnt < AMREX_SPACEDIM; cnt++) {
+      slp[cnt] = (xp[cnt] - x_low[cnt]) / dx[cnt];
+    }
 
-  const int XDIR = 0;
-  const int YDIR = 1;
-  const int ZDIR = 2;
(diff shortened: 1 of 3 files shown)


Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,324 @@
...
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> x_low,
+  amrex::Array3D<amrex::Real, 0, 1, 0, 1, 0, 1> cell_data,
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> dx)
+{
+  amrex::Real alpha, beta, gama;
+  amrex::Real value = 0.0;
+
+  alpha = 0.0;
+  beta = 0.0;
+  gama = 0.0;
```
marchdf: all this should be defined as one line. And use ternary expressions for beta and (renamed?) gamma.

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+  const int XDIR = 0;
+  const int YDIR = 1;
+  const int ZDIR = 2;
```
marchdf: I think these exist in other place. Also you can see a trilinear interpolation (with a for loop) here: https://github.com/AMReX-Combustion/PeleC/blob/development/Exec/RegTests/HIT/prob.H#L67

marchdf (commented):
My only comment really is that this feels complicated. We do probes in amr-wind with particles and it is fast and very scalable. I would have recommended that route before but you are quite far along here so if this works, maybe it's fine. My recommendation would be to try a test with millions of probes (users are going to do that to you) and make sure you don't spend too much time in init and timestepping.

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,324 @@
...
+        if (m_interpType_d == Linear) {
+
+#if (AMREX_SPACEDIM == 1)
+          {
+            cell_data_d(0, 0, 0) = state(low_cell_idx_d[0], 0, 0, stIdx);
+            cell_data_d(1, 0, 0) = state(low_cell_idx_d[0] + 1, 0, 0, stIdx);
+          }
+#elif (AMREX_SPACEDIM == 2)
+	        {
+	        	cell_data_d(0, 0, 0) = state(low_cell_idx_d[0],       low_cell_idx_d[1],       0, stIdx);
```
baperry2: This file looks like it need to be clang-formatted

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,324 @@
...
+  alpha = 0.0;
+  beta = 0.0;
+  gama = 0.0;
+
+  const int XDIR = 0;
+  const int YDIR = 1;
+  const int ZDIR = 2;
+
+  alpha = (xp[XDIR] - x_low[XDIR]) / dx[XDIR];
+  if (AMREX_SPACEDIM >= 2) {
```
baperry2: do these using the `AMReX_D_TERM` macro rather than runtime conditionals

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,324 @@
...
+#include <regex>
+#include <cstdio>
+
+AMREX_GPU_DEVICE
+AMREX_FORCE_INLINE
+amrex::Real
+LinearInterpolate(
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> xp,
+  amrex::GpuArray<amrex::Real, AMREX_SPACEDIM> x_low,
+  amrex::Array3D<amrex::Real, 0, 1, 0, 1, 0, 1> cell_data,
```
baperry2: Change the variable name from `cell_data`, which might lead to confusion because there is a different AMReX data structure called cellData. But also, instead of creating the new cell_data structure and copying data into it, why not just pass the original array4 and the relevant index in?

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,324 @@
...
+
+  alpha = (xp[XDIR] - x_low[XDIR]) / dx[XDIR];
+  if (AMREX_SPACEDIM >= 2) {
+    beta = (xp[YDIR] - x_low[YDIR]) / dx[YDIR];
+  }
+  if (AMREX_SPACEDIM == 3) {
+    gama = (xp[ZDIR] - x_low[ZDIR]) / dx[ZDIR];
+  }
+
+  value += (1.0 - alpha) * (1 - beta) * (1 - gama) * cell_data(0, 0, 0);
```
baperry2: add some dimensionality checks here so the unused pieces for lower dimensions don't get called

baperry2 (commented):
As @marchdf says, this won't scale for very large numbers of probes, but I'm fine with that for now.

Source/Utility/Diagnostics/DiagProbe.cpp
SreejithNREL: Removed the variables alpha, beta and gama and replaced with an array from the code snippet you suggested.

Source/Utility/Diagnostics/DiagProbe.cpp
SreejithNREL: removed XDIR, YDIR and ZDIR and added a loop to populated the values in the slp array

Source/Utility/Diagnostics/DiagProbe.cpp
SreejithNREL: I tried clang tidying again, but nothing seems to change. Are you sure?

Source/Utility/Diagnostics/DiagProbe.cpp
SreejithNREL: Replaced the code snippet with an array that Marc suggested. Now we don't need these conditions

Source/Utility/Diagnostics/DiagProbe.cpp
SreejithNREL: Renamed the variable name.
Do you think changing it to array4 can get some benefits? If not, I would like to keep it this way and avoid rewriting the code once again

Source/Utility/Diagnostics/DiagProbe.cpp
SreejithNREL: For low-dimensional cases, the third index values in cell_data will be zeros, contributing zero values to the interpolation.

SreejithNREL:
Thanks Marc and Bruce for your suggestions. I did think about using particles as probes, but I was under the impression that including particle functionalities (say in an otherwise non-particle simulation) might cause more overheads. Also, tbh, I didn't foresee anything more than 10-20 probes (max) to be used in practical simulations. Maybe, we can develop the same functionality using particles in the future?

Source/Utility/Diagnostics/DiagProbe.cpp
baperry2: Sure, maybe the code could be a bit simpler by not creating the intermediate data structure, but if what's there works on GPU it should be fine and not worth rewriting.

Source/Utility/Diagnostics/DiagProbe.cpp
baperry2: Looks like clang-format isn't touching the `#elif` and `#else` blocks inside of the ParallelFor for some reason. Perhaps a bug in clang-format. In any event, I got the file properly formatted by temporarily commenting out these directives and running clang-format.

baperry2 (approved):
I agree, let's leave this implementation for now and consider a particle based implementation in the future.

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,328 @@
...
+            state(m_probe_idx_d[0], m_probe_idx_d[1], m_probe_idx_d[2], stIdx);
+#endif
+        }
+      });
+  }
+
+  amrex::ParallelDescriptor::ReduceRealSum(
+    m_values_at_probe_d.data(), static_cast<int>(m_values_at_probe_d.size()));
+
+  amrex::Gpu::copy(
```
marchdf, line 315: Only IO rank needs to do this copy right?

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,328 @@
...
+          tmp_values_d[n] = state(m_probe_idx_d[0], m_probe_idx_d[1], 0, stIdx);
+#else
+          tmp_values_d[n] =
+            state(m_probe_idx_d[0], m_probe_idx_d[1], m_probe_idx_d[2], stIdx);
+#endif
+        }
+      });
+  }
+
+  amrex::ParallelDescriptor::ReduceRealSum(
```
marchdf, line 312: this is the kinda thing that is crushing users in amr-wind (until I fixed it ;) ). Each rank is holding way more data than it actually needs. Just a comment, not something to fix here.

Source/Utility/Diagnostics/DiagProbe.cpp
```diff
@@ -0,0 +1,328 @@
...
+    auto& dx_finest_lev_probe_d = dx_finest_lev_probe;
+    auto& m_interpType_d = m_interpType;
+
+    amrex::ParallelFor(
+      m_fieldIndices_d.size(), [=] AMREX_GPU_DEVICE(int n) noexcept {
+        amrex::Array3D<amrex::Real, 0, 1, 0, 1, 0, 1> neighbour_cells{
+          0.0}; // Neighbour cell solution values
+        int stIdx = idx_d_p[n];
+
+        if (m_interpType_d == Linear) {
```
marchdf, line 259: I would pull this if out of the parallelfor and have 2 parallelfors. This is to avoid potential thread divergence on the GPU. There's very little shared code so I think it would be easy to do.

PR #545, opened 2024-11-07
Tidied up nested namespace in Utilities for clang-tidy

d-montgomery:
The most recent commit failed to pass the clang-tidy test in PeleLMeX.  This PR should fix the problem.  Also, I removed Utilities.H from PelePhysics.H per @baperry2's request.  To include unit conversions in PeleLMeX or C we'll need to add Utilities.H to the associated PeleLMeX.H (or PeleC.H) headers as is done with the other utilities.  This works in my local PeleLMeX branch.

### DIFF
--- a/Source/PelePhysics.H
+++ b/Source/PelePhysics.H
@@ -7,7 +7,6 @@
 #include "PelePhysicsConstraints.H"
 #include "EOS.H"
 #include "Transport.H"
-#include "Utilities.H"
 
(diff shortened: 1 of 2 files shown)


baperry2:
Can you update your PeleLMeX pull request to use this version of PelePhysics so we can verify that this resolves the clang-tidy issues before merging?

baperry2 (approved):

PR #549, opened 2024-11-20
CEPTR: PLOG reaction support at constant pressure

### LINKED ISSUE #538: Feature Request: Plog reaction support in CEPTR (currently leads to AttributeError: 'cantera.reaction.PlogRate' object has no ...)
Hello,It seems that cepter does not support the cantera type: pressure-dependent Arrhenius form of chemical reaction.which like that :
- equation: C6H5 + O2 => CJVCCVCCVO + CO  # Reaction 854
  type: pressure-dependent-Arrhenius
  rate-constants:
  - {P: 0.1 atm, A: 1.128e+10, b: 0.64, Ea: -5135.0}
  - {P: 1.0 atm, A: 1.02e+14, b: -0.44, Ea: -1650.0}
  - {P: 10.0 atm, A: 2.42e+27, b: -4.09, Ea: 1.038e+04}
  - {P: 100.0 atm, A: 9.8e+43, b: -8.56, Ea: 2.683e+04}

when I try to use CEPTR convert(use convert.sh) yaml to  generate C++ mechanism files ,it ouccurs that 

> Traceback (most recent call last):
>   File "<string>", line 1, in <module>
>   File "/home/zhangjiaqian/Desktop/PeleLMeX/Submodules/PelePhysics/Support/ceptr/ceptr/ceptr.py", line 201, in main
>     convert(
>   File "/home/zhangjiaqian/Desktop/PeleLMeX/Submodules/PelePhysics/Support/ceptr/ceptr/ceptr.py", line 71, in convert
>     conv.writer()
>   File "/home/zhangjiaqian/Desktop/PeleLMeX/Submodules/PelePhysics/Support/ceptr/ceptr/converter.py", line 453, in writer
>     cp.production_rate(
>   File "/home/zhangjiaqian/Desktop/PeleLMeX/Submodules/PelePhysics/Support/ceptr/ceptr/production.py", line 199, in production_rate
>     pef = (reaction.rate.pre_exponential_factor * ctuc).to_base_units()
>            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> AttributeError: 'cantera.reaction.PlogRate' object has no attribute 'pre_exponential_factor'
what i use mechanism.yaml file  url is below.
(https://github.com/grabtheNS/soot-chemical-mechanism-/blob/78840684f7a85712898c1093095bc2be4986697b/kaust.yaml)

It does not seem to support the pressure dependent chemical reaction format yaml file, but I am not sure what is the cause of this error.
Best regards!

terencelehmann:
Suggestion for #538 at constant pressure, which is provided by the user when converting the mechanism.

Changes to code:
- ceptr/production.py : Assessing if reaction type is "pressure-dependent-Arrhenius", and call utilities function for rate calculation
- ceptr/jacobian.py : Assessing if reaction type is "pressure-dependent-Arrhenius", and call utilities function for rate calculation
- ceptr/utilities.py: New function evaluate_plog -> When called for the first time, gets pressure from used and stores in global variable for reuse. Then interpolation of rate coefficients.

Tests:
- Generation of mechanisms within and outside the PLOG pressure limit. Checked computed rate coefficients and units.

Possible improvements:
- The current version only supports constant pressure, since the rate coefficients are evaluated at the time of compilation. As such it is only a solution for PeleLMx and removes the manual replacement of PLOG reactions. A warning for this is implemented. However, as mentioned in #487 
 ([https://github.com/AMReX-Combustion/PelePhysics/discussions/487#discussioncomment-8976018](url)) a check might be even better. However, I do not know how to do that.


marchdf:
Thank you for doing this! Do you have a small mech that you could add to the PR as and example?

baperry2:
Thanks for the submission @terencelehmann 

For Discussion: 
- I like this capability, but I worry about the potential for people to accidentally use mechanisms that get generated though this process for the wrong pressure. One solution is that for mechanisms with plogs, we create a subdirectory with the specific pressure in the name to store `mechanism.{h,cpp}`. For example something like `Mechanisms/POLIMI2020/p1atm/mechanism.{h,cpp}`. The current gmake build system would work by specifying `Chemistry_Model = POLIMI2020/p1atm`, and this would ensure users are aware that they are using a mechanism for that specific pressure. It would also allow keeping around multiple versions for multiple pressures. Not sure how hard this would be to implement in CEPTR. @malihass @marchdf thoughts?
- We should also add the new mechanism to `list_mech` so it gets included in batched mechanism generation, which will require some way of specifying the pressure a priori rather than on the fly on the command line.

malihass:
I was going to suggest indicating the pressure in a `.toml` input like we do for the symbolic Jacobin stuff, but I prefer your idea Bruce, I can see this being a recipe for mistakes if we don't explicitly write the value of pressure next to the mechanism name. I don't anticipate it being too difficult to implement in ceptr.

malihass:
Also I ran cases with 0.5 atm, 1atm 2atm and everything looks on par with Cantera, with numerical Jac and Analytical Jac, great job!
plogval

terencelehmann:
Thank you @baperry2 @malihass for the input! I'm happy to implement these suggestions (next week). Where exactly is the list_mech file used?

baperry2:
`list_mech` is just a list of all working non-QSSA mechanisms in PelePhysics. We use it to regenerate all mechanism code in  batched manner when necessary:

> $ cd ${PELE_PHYSICS_HOME}/Support/ceptr
> $ poetry run convert -l ${PELE_PHYSICS_HOME}/Mechanisms/list_mech

The catch is this won't work for specifying the pressure through an input prompt as its set up now. Ideally we'd have a way to specify the pressure for mechanisms with plogs as a command line argument when running this command or perhaps put that information into the list_mech file and parse it from there so it could be specified for mechanisms individually.

terencelehmann:
I have now fixed the issues mentioned above. Also, in this new version, I have used the pressure attribute in the Cantera mechanism object /Solution class to store the pressure rather than in a global variable in ceptr.utilities.
Let me know if there is anything else to fix.

baperry2:
Looks good. Thanks for the updates - I like getting rid of global variables as well. I'd like to add some documentation for PLOGs in CEPTR: https://github.com/ITV-RWTH/PelePhysics/pull/10

If that documentation looks good, I think we can merge this.

malihass (approved):

baperry2 (approved):
Thanks again for the contribution!

PR #560, opened 2025-02-14
Inconsistency of MW/iMW precisions in CEPTR

EnnaDelfen:
With this PR, the precision of the molecular weight written by CEPTR matches the one specified for the inverse of the molecular weights (.16f). 
This fixes the inconsistency observed when multiplying the MW and the IMW for electrons, where that ratio was too far from 1 (to less than machine precision, which caused problems). 
NOTE: one could also choose to reduce the precision of IMW instead, thus not breaking all unit tests... but it seems odd to limit the precision to less than machine precision. 
NOTE 2: There is still an inconsistency with transport values, for instance with egtransetWT dumped by transport.py, where the precision here is .8E . Not sure what we may want to do here.


marchdf:
Thanks for doing this! To save you some hassle I was going to push some formatting changes and then apply what you did to all the mechs. Can you give me permission to push to this branch or do those things on your end?

baperry2:
How about using `.16e` as the format instead of `.16f`, so we always keep to machine precision regardless of order of magnitude of the quantities?

EnnaDelfen:
Hi Marc (and Bruce), sure go ahead! Let me know if you need me to do anything else.
Edit: just realizing I need to give you permission most likely, let me figure that out ...

marchdf:
Tried to push and got 
```
ERROR: Permission to PlaFCo/PelePhysics.git denied to marchdf.
fatal: Could not read from remote repository.
```
Can you fix that @EnnaDelfen ?

EnnaDelfen:
I should have given you access..

marchdf:
Ok I switched to 16e. And I fixed qss regen path for the plog stuff.

esclapez (approved):
Nice ! I'm not sure we really know molecular weights down to that much precision, but at least it's consistent throughout CEPTR.

PR #566, opened 2025-04-13
Add functionality to use non-periodic inflow data

drummerdoc:
If the turbulence inlet data HDR indicates that the file is not periodic in the k direction, activate a second mode of interpolating the data in the file.  In this case, each plane of data has an associated timestamp, and the parabolic interpolation in the k direction allows nonuniform spacing associated with the unequal time intervals between planes.  The HDR file describing the data must be augmented with the timestamp for each plane, and the "isswirltype" flag within the turbInflow class is turned on.  To implement this capability, minor changes were made to the logic of the interpolation even for periodic data, but the changes should not impact the results in those cases.  

Notes:
1. Operating in this mode is considerably less robust.  The time value input to the interpolation function must lay within the bounds of the times provided in the input file or an assertion failure is triggered.
2. A more thorough documentation of this capability needs to be added to the Pele instructions, as this significantly complicates usage and will like lead to some confusion, especially given the other methods supported for inflow fields
3. A utility to create these non-periodic turbulent files will be added to AMReX in a separate pull request [4418](https://github.com/AMReX-Codes/amrex/pull/4418), and a functionality has been added to PeleLMeX [502](https://github.com/AMReX-Combustion/PeleLMeX/pull/502) to generate the planes of data that are input to that utility.  Thus, there are 3 separate components to this capability that are required for its use.


baperry2 (approved):
Fixed some bugs and improved the errors to be more helpful to the user. Still a lot going wrong with turbinflow in general (e.g., #571), but I'm going to merge this piece

PR #567, opened 2025-04-25
add 4spec nonreacting mech for co2snow

baperry2:
4 nonreacting species truncated from DRM19. Useful for calculations involving flue gas.


d-montgomery (approved):
Thanks for adding this mechanism.

PR #568, opened 2025-04-29
update amrex and sundials to latest releases

baperry2:
In preparation for making a release of PelePhysics


baperry2:
Sundials beyond 7.2.0 seems to have a compatibility issue with AMReX in CMake, so removing the sundials update from this PR.

https://github.com/baperry2/PeleC/actions/runs/14740769324

baperry2:
We can update to Sundials 7.3.0 but that will require adding `set(EXAMPLES_INSTALL OFF)` in `CMake/SetSundialsOptions.cmake` for PeleC/LMeX and addressing some deprecated functions, so I'll wait for a later release to do that.

PR #571, opened 2025-05-08
turbinflow fix for off by 1/2 cell issue

baperry2:
Attempting a very coarse grid turbinflow on a fine grid PeleLMeX simulation reveals issues in the way we are apply the turbinflow. Because we use `round` rather than `floor`, our values for the normalized location being interpolated range from -0.5 to 0.5 instead of 0 to 1. This means the turbinflow is 1/2 turbinflow cell displaced from where it should be. This also messes up interpolation because negative values correspond to extrapolating beyond the left edge of our 3-cell interpolation stencil and also there is a discontinuity at the half cell spacing. 

I believe these modifications result in the currently intended behavior of turbinflow: in the end, turbulence from a triply periodic turbulence cube  of cell-centered data gets applied in the desired location, but the turbulence is offset by dx/2 relative to the initial data.

Test case: PeleLMeX on a -2pi to 2pi domain with 256 points in each direction, applying a turbinflow with 2pi periodicity and a 4x4 base grid.

Current behavior:
Screenshot 2025-05-08 at 5 35 07 PM

Corrected behavior:
Screenshot 2025-05-08 at 5 37 20 PM

### DIFF
--- a/Source/Utility/TurbInflow/turbinflow.cpp
+++ b/Source/Utility/TurbInflow/turbinflow.cpp
@@ -65,7 +65,9 @@ TurbInflow::init(amrex::Geometry const& /*geom*/)
       }
 
       pp.query("turb_nplane", tp[n].nplane);
-      AMREX_ASSERT(tp[n].nplane > 0);
+      AMREX_ASSERT(
+        tp[n].nplane >
+        3); // need at least 4 planes for quadratic interpolation + 1 extra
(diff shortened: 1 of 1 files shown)


drummerdoc:
The +1 -> +2 change is surprising...but I assume correct.  Need to doodle somewhere on scratch paper to understand that one.

drummerdoc (approved):

baperry2:
I'm actually not sure on the +1 to +2 change so I will double check.

PR #572, opened 2025-05-09
enable dumping the 1st plane of ghost cells with DiagFramePlane

baperry2:
Currently if the requested plane location for the DiagFramePlane Diagnostic is not available, we just silently do nothing. After this PR, we will raise an error, unless the user has specified the `dump_ghost_if_OOB` option, in which case the first ghost cell in the relevant face will be dumped (all levels get their respective first ghost cell). This is intended as debugging capability, for example giving users direct access to visualize what bcnormal is doing. 

In PeleLMeX, the first ghost cell on domain boundaries corresponds to the BC at the boundary **face**.

In PeleC, the first ghost cell on domain boundaries corresponds to the state 1/2 cell over from boundary face, which gets set such that applying standard operators will lead to the desired values on the face.

### DIFF
--- a/Docs/sphinx/Utility.rst
+++ b/Docs/sphinx/Utility.rst
@@ -132,7 +132,7 @@ The following provide examples for each diagnostic in PeleLMeX (in PeleC, all di
     peleLM.xnormP.interpolation = Linear                            # [OPT, DEF=Linear] Interpolation type : Linear or Quadratic
     peleLM.xnormP.field_names = x_velocity mag_vort density         # List of variables outputted to the 2D pltfile
     peleLM.xnormP.n_files = 2                                       # [OPT, DEF="min(256,NProcs)"] Number of files to write per level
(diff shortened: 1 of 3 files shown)


malihass (changes requested):
Thanks for that, it works great in a turbulent channel flow case!
One change that would be good
Can we dump the error message or the info message only once (not MPI process times)?

Also (not for this PR), but it the LMeX doc, it would be nice to have the dump_ghost_if_OOB option.

baperry2:
Modified so the warning only gets printed once. Leaving the error as is, because we want to ensure an error is always raised if needed.

Moved Docs here so we don't have to maintain separate Docs for this capability in LMeX and C, and added something for the new flag.

malihass (approved):
Super! Thanks for the change
Yesterday we noticed that the plane at Step 0 was not dumped, so if you want to add this that would be great, otherwise it is already great the way it is!

baperry2:
The change to dump the plane at step 0 will need to go in PeleC/LMeX I think

PR #576, opened 2025-05-20
Mechanisms for Jet-A and non-reacting POSF fuels

d-montgomery:
This PR includes two mechanisms for Jet-A (posf10325 and posf10325_qss), which were originally converted by @baperry2 [here](https://github.com/baperry2/PelePhysics/tree/new-mechanisms/Mechanisms). A third non-reacting mechanism (posf_nonreacting) is included for fuels A1, A2, A3, and C1, with O2, N2, CO2, and H2O.

All three are based on the [HyChem mechanism from Stanford](https://web.stanford.edu/group/haiwanglab/HyChem/pages/download.html).


baperry2:
@dmontgomeryNREL: can you rerun the CEPTR conversion from yaml files for the posf10325 and posf10325_qss mechanisms and verify nothing has changed? Because I ran the conversion a while ago just want to make sure nothing has changed for them

d-montgomery:
> @dmontgomeryNREL: can you rerun the CEPTR conversion from yaml files for the posf10325 and posf10325_qss mechanisms and verify nothing has changed? Because I ran the conversion a while ago just want to make sure nothing has changed for them

Looks like the formatting changed slightly with additional sig figs.

baperry2:
Last thing is to add the mechanism names to `Mechanisms/list_mech` and `Mechanisms/list_qss_mech` so they get regenerated as needed during the batched mechanism generation process

baperry2 (approved):

PR #580, opened 2025-06-02
widen tolerance for grid checking in turb_box inflow generator

baperry2:
Grids that are actually uniform are raising an error because the current tolerance `1-e14` can be a bit to tight for determining what constitutes equal with floating point precision in the relevant calculations. Therefore, widen the tolerance to `1e-12`


d-montgomery:
Is this consistent with the AMReX requirements for a uniform grid, specifically for cases with EB?

baperry2:
This is mostly unrelated to that. We extract x,y,z,velocity data from a data file generated by a python script (which assumes a uniform grid) and map that into a FAB (which by definition has a uniform grid). The check is just to make sure that users don't abuse this tool to try to process data from other sources that didn't have a uniform grid.

d-montgomery (approved):

PR #583, opened 2025-06-04
Allow 0 or negative turb conv vel in turbinflow

baperry2:
There's no need to enforce postive turbulent convective velocity. Users may want a frozen inflow condition or even to march through an inflow file in reverse.


SreejithNA:
That's right. I don't think such a restriction is necessary.

SreejithNREL (approved):

PR #586, opened 2025-06-25
Update MAGMA to 2.9.0

### LINKED ISSUE #585: Updating MAGMA
When compiling MAGMA for HIP with gfx90a architecture, some of the architecture flags aren't passed on in version 2.7.1 of MAGMA, resulting in some of the files being compiled for some default architecture (gfx803) and throwing a `hipErrorNoBinaryForGpu` error. This issue was noticed [here](https://github.com/icl-utk-edu/magma/issues/11) and subsequently fixed, but I needed to update the MAGMA version [here](https://github.com/AMReX-Combustion/PelePhysics/blob/1fa488ec42e994936402991dca14c68c39bbe2a8/ThirdParty/GNUmakefile#L10) and everything worked again. Do we foresee any issues with just updating this value to 2.9.0 by default?

ThomasHowarth:
As the title says; resolves #585

### DIFF
--- a/ThirdParty/GNUmakefile
+++ b/ThirdParty/GNUmakefile
@@ -7,7 +7,7 @@ CUDA_ARCH       ?= 70
 AMD_ARCH        ?= gfx90a
 
 SS_VERSION=5.4.0
-MAGMA_VERSION=2.7.1
+MAGMA_VERSION=2.9.0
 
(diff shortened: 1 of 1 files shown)


baperry2 (approved):
Thanks, it turns out that @SreejithNREL was running into the same issue recently on Frontier and this fix resolves it.

PR #587, opened 2025-06-25
use amd libsci with hip compilers for magma

baperry2:
On Frontier with PrgEnv-amd, we need `libsci_amd.so` instead of `libsci_cray.so` when compiling MAGMA with `COMP=hip`

@ThomasHowarth - I encountered this when checking #585. If this breaks things for you, I can try to make the check I added machine-specific.


ThomasHowarth (approved):
This all worked on the machine I'm using, also with PrgEnv-amd

jrood-nrel (approved):

PR #588, opened 2025-06-26
Add turbulent forcing functionality to generate and maintain homogeneous isotropic turbulence.

efhunt:
This PR adds functionality to generate and maintain homogeneous isotropic turbulence. The method used is the same as that used in PeleLM and IAMR (Andrew Aspden's forcing, see Aspden et al (CAMCoS 2008)). 

A separate PR is being submitted to PeleLMeX which will add the appropriate hooks in addition to an example configuration.

For details on the implementation see Docs/sphinx/Utility.rst section "Forced Turbulence Implementation Details".

This was the combined effort between Andrew Aspden, Thomas Howarth and Edward Hunt where we stated with Candace Gilet implementation in IAMR. 

The code has been tested using PeleLMeX on CPUs (by myself gcc (Debian 12.2.0-14+deb12u1) 12.2.0)) and Max Schneider on GPUs (ml Stages/2024 NVHPC/25.1-CUDA-12 OpenMPI/4.1.5 CMake/3.26.3)


Docs/sphinx/Utility.rst
```diff
@@ -136,6 +137,50 @@ inflow plane data and interpolation approaches for periodic and nonperiodic tang
...
+For ff_factor, nmodes, forcing_epsilon, spectrum_type and moderate_zero_modes, it is best to leave these as defaults unless you are confident on
+the consiquences.
+
+Forced Turbulence Implementation Details
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+The strategy is to inject energy into the large scales using a source term in the momentum equation; the turbulence then develops naturally through the
+Richardson-Kolmogorov cascade.  The original description is in Aspden et al (CAMCoS 2008); a few modifications (e.g. tweaking the amplitudes of the
+modes through the random seed, which reduced the temporal variation in u_rms) and explicitly using a divergence-free form (mainly to keep Minion happy;
+see Almgren et al SIAM 2013).  We typically run in a cube, or 4:1 box; larger boxes can achieved using the hack_lz option, which uses a periodic
```
baperry2: Also include the DOI links to the referenced papers here. I would omit the commentary "to keep Minion happy" and just leave the paper reference if it explains the point.

Source/Utility/TurbForcing/TurbForcing.cpp
```diff
@@ -0,0 +1,799 @@
...
+          FTX(kx, ky, kz) =
+            (m_tfp.m_freqMin + m_tfp.m_freqDiff * mersenne_twister::Random()) *
+            TwoPi;
+          mersenne_twister::Random(); // dummy FTY (don't remove)
+          mersenne_twister::Random(); // dummy FTZ (don't remove)
+          // Translation angles, theta=0..2Pi and phi=0..Pi
+          TAT(kx, ky, kz) = mersenne_twister::Random() * TwoPi;
+          mersenne_twister::Random(); // dummy TAP (don't remove)
+          // Phases
+          FPX(kx, ky, kz) = mersenne_twister::Random() * TwoPi;
```
baperry2, line 203: Could all these be replaced with `amrex::Random()` to simplify the code and eliminate the need for the MersenneTwister code?

Source/Utility/TurbForcing/TurbForcing.cpp
```diff
@@ -0,0 +1,799 @@
...
+  // target velocity fluctuation
+  pp.query("urms", m_tfp.m_urms);
+  // time offset (e.g. for starting from precursor turb sim)
+  pp.query("time_offset", m_tfp.m_time_offset);
+  // allow peridodic reproduction in z
+  // can be used as a flag i.e. =1 for a factor 2
+  // or as the factor itself
+  pp.query("hack_lz", m_tfp.m_hack_lz);
+
+  pp.query("rho_incompressible", m_tfp.m_rho_incompressible);
```
baperry2: It seems bug-prone to require that the user specify another rho_incompressible in addition to what they specify for the base solver. How about passing the incompressible flag and rho (for incompressible simulations) to the init function, so they can be specified directly by PeleLMeX?

Source/Utility/TurbForcing/TurbForcing.cpp
ThomasHowarth: @ajaspden is this possible, or is the amrex::Random() "too" random?

ajaspden: Replacing the random number generator will change the resulting forcing term, which will then not give the same answer.  It's suboptimal, but the whole forcing was made to work with this Mersenne Twister around the dawn of time and hasn't changed much since.  It would be possible (and desirable) to use amrex::Random(), but the u_rms would be off, and it would be a considerable effort to recalibrate.

Docs/sphinx/Utility.rst
efhunt: Thanks. Amended.

Source/Utility/TurbForcing/TurbForcing.cpp
baperry2: I'm not familiar with the inner workings of this forcing approach, but I'm curious why recalibration would be needed if both are drawing uniformly distributed pseudo random numbers? Shouldn't the resulting forcing be statistically equivalent?

I am ok with leaving the Mersenne Twister stuff in for now if needed though.

ajaspden: It's a good question.  Basically, because we only force a small number of modes at low wavenumbers, there aren't enough modes to guarantee that it's statistically isotropic, especially since the large modes are so dominant.  I would anticipate a better approach could be found to choose these numbers (such as taking all modes to be the same amplitude (in 3d wavenumber space), and allowing the temporal variation and phase shift to take care of the rest), but it would take some playing to make sure it gives something sensible.

Source/Utility/TurbForcing/TurbForcing.cpp
efhunt: Hi, thank you for your review and comments. I think this is a good idea. I am unlikely to get round to making the changes this week so will aim to have it pushed on Monday.

efhunt: Hi baperry2. I have made the changes. Hopefully what I pushed was what you had in mind (required changes have also been made to LMeX).

baperry2: Looks good.

baperry2 (approved):
Thanks again for adding the turbulence forcing capability!

PR #589, opened 2025-06-27
correct tmax and tmin in temp_given_ey,hy based on mechanisms bounds

EnnaDelfen:
With this commit, the functions GET_T_GIVEN_EY and GET_T_GIVEN_HY will effectively use the most stringent prescribed thermo bounds for the loop over the temperature, and extrapolate outside these bounds. It's basically restoring what the comments are saying "max lower bound for thermo def" and "min upper bound for thermo def" .
It may change results quite significantly however, particularly the lower bound at 90 was quite low and will most likely always be overwritten. Is it best to extrapolate or to fetch a value with a polynomial that isn't supposed to be used over this range? Not sure.
The reason for the proposed modif in my case concerns the higher bound: we sometimes have poly fits that run higher than 4000, so a conservative change could be to change that upper bound only if we have a mechanism that allows it


baperry2:
This change sounds good to me. Can you regenerate the mechanisms based on these changes?

EnnaDelfen:
I'm changing the regular mechs, I've tried the QSS ones but they give me very different results in the Jacobian and I'm not sure if that's completely normal? Maybe someone with better understanding of these ceptr routines could have a second look at it (alternatively, if there's a specific test I can try I could do that to make sure it's all OK)

baperry2:
@malihass or @marchdf could one of you take a look?

marchdf:
Yup, looking now.

marchdf:
The diffs in QSS are not due to any changes in this PR. Rerunning the code generation on the `development` branch generates the same diffs. I pushed the updated QSS to this branch.

I think this is just due to how the sympy is simplifying the expressions. I tried tracking the difference through but quickly got lost.

marchdf (approved):

baperry2 (approved):

baperry2:
I believe these changes are causing small diffs in both PeleC and PeleLMeX tests. I don't think that is unexpected, but will verify.

https://github.com/AMReX-Combustion/PeleC/actions/runs/15995545253/job/45118031526?pr=961

https://github.com/baperry2/PeleLMeX/actions/runs/15989264479/job/45099282203

EnnaDelfen:
Yes, it should change results. As I mentioned in my first message, many mechanisms had poly fits valid for a smaller range than the one that was hard coded (90-4000), so you're now effectively directly interpolating instead of fetching a polynomial value... Both options are "wrong", but which one is the most wrong remains to be investigated

baperry2:
Yep, the cases that have small diffs appear to be when the temperature is slightly out of the bounds (e.g. a minimum of 298 when the lower bound is 300). Things are working as expected.

Issue #579, opened 2025-06-02
Bug Report: Incorrect Handling of Third-Body Concentration and Stoichiometry in Falloff Reaction Implementation


bssoriano:
The current implementation of unimolecular/recombination falloff reactions incorrectly uses the total mixture concentration as the third-body collider and includes third-body species in the stoichiometric concentration products for forward and reverse rates. This leads to incorrect rate calculations, especially in mechanisms where only specific species enhance the reaction.

Reaction Example (Chemkin Format):

```
H + O2 (+H2O) ⇌ HO2 (+H2O)   2.15E12  0.292  -592.0  
LOW / 2.28e+19 -0.977  0.0 /  
TROE / 1.0 1.0E-10 1.0E30 1.0E30 /
```

---

When this reaction is parsed from the YAML file to C++, the rate expressions become:

```cpp
Corr = mixture;
qf = Corr * k_f * (sc[H] * sc[O2] * sc[H2O]);  
qr = Corr * k_f * exp(-(g_RT[H] + g_RT[O2] - g_RT[HO2])) * refC * (sc[H2O] * sc[HO2]);  
```

#### 1. Incorrect Third-Body Concentration
The code uses the total mixture concentration (`[mixture]`) as the falloff collider concentration:
```cpp
Corr = mixture;
```

This is incorrect. The reaction above specifies H₂O as the only third-body collider. The effective third-body concentration should be:

`Corr = sc[H2O];  // or alpha_H2O * sc[H2O] if efficiencies are specified`

2. Incorrect Stoichiometry in Rate Expressions

Third-body species like H₂O should not appear in the rate expressions for forward or reverse reactions since they are not consumed or produced in the reaction:

```
Corr = F * F_troe;
qf = Corr * k_f * (sc[H] * sc[O2] * sc[H2O]); // ❌ Incorrect
qr = Corr * k_f * exp(-(g_RT[H] + g_RT[O2] - g_RT[HO2])) * refC * (sc[H2O] * sc[HO2]);   // ❌ Incorrect
```
The correct rate expressions should be:

```
Corr = F * F_troe;
qf = Corr * k_f * sc[H] * sc[O2]; // ✅ Correct
qr = Corr * k_f * exp(...) * sc[HO2]; // ✅ Correct
```

⚠️ Additional Note
[...]

PR #590, opened 2025-06-27
Ceptr: Extend PLOG feature and PMF generator bugfix

terencelehmann:
Small pull request for an extension of the PLOG feature of ceptr to "three-body-pressure-dependent-Arrhenius" reactions. 

Additionally, this PR fixes a bug in the PMF generator for PLOG mechanisms. Since the mechanism file name is taken from the list_mech file, the --plog option needs to be removed to find the file.


baperry2:
While waiting on the mechanism to be published, you can follow [these steps](https://amrex-combustion.github.io/PelePhysics/DeveloperGuide.html#python) to get all the tests to pass

ThomasHowarth:
FYI - mechanism can be found [here](https://www.sciencedirect.com/science/article/pii/S001021802500447X).

baperry2:
Am I missing something or does this mechanism not actually include three-body PLOGs? When I run the ceptr conversion, it doesn't look like the `three-body-pressure-dependent-Arrhenius` conditional ever gets activated. Maybe they were in the full but not the lite version of the mechanism?

Either way it's good capability to have and a good mechanism to include with the code.

baperry2 (approved):
The issue I noted above is that `reaction.rate.type == "three-body-pressure-dependent-Arrhenius"` is never true because here `three-body` is a subtype of `pressure-dependent-Arrhenius`. Doing the check on `reaction.reaction_type` instead of `reaction.rate.type` would resolve the issue. However, the current code functions correctly because the reactions still get caught as `pressure-dependent-Arrhenius` and `plog` is set to `True`.

Additionally, this mechanism has the issues with certain 3 body falloff reactions outlined in #579 that result in small differences in predictions vs. Cantera.

With these notes, I am going to merge this PR so that the fixes are added to the code and the mechanism included can more easily be used for testing associated with #579.

PR #592, opened 2025-07-02
Split vapor phase dep species across its liquid specs when computing skin composition

baperry2:
With spray, multiple liquid species can be set to contribute mass to a single gas phase species. 

Currently, when computing the skin compositions with the 1/3 rule, the gas phase composition that we use is the full mass fraction of the gas phase dependent species for all liquid species. As a result, having a 50/50 droplet of two identical species evaporating into a single species does not give the same vaporization as a droplet that is 100% of a single species.

Instead, we can split the gas concentration of the gas phase species across the relevant liquid species in proportion to the equilibrium composition. This is somewhat arbitrary, but so is the idea of having multiple liquid species evaporating into a single gas phase species.

The implementation is a bit inefficient (looping over SPRAY_FUEL_NUM twice) but more efficient approaches would require more complex data structures communicated between host/device and the computational cost should be negligible anyway.

@dmontgomeryNREL - I know you'll be out of pocket for a while, but there's no rush on reviewing this.


baperry2:
To test this, I created a mechanism with 3 identical heptane species: `NC7H16{A,B,C}`. Two cases are run: one with 100% NC7H16A in the liquid phase contributing to NC7H16A in the has phase; and one with 50/50 A/B in the liquid phase both contributing to NC7H16C in the gas phase. Because all species are identical, we should expect that the amount of NC7H16A in the gas phase in Case 1 equals the amount of NC7H16C in the gas phase in Case 2. We see significant differences for the current code, but only a very small difference using this PR (I'm not sure what's causing the very small remaining difference).

Results of Case 1 (unchanged due to this PR):
Screenshot 2025-07-14 at 1 53 59 PM

Results of Case 2 (prior to this PR):
Screenshot 2025-07-14 at 1 54 14 PM

Results of Case 2 (after this PR):
Screenshot 2025-07-14 at 1 54 06 PM

To reproduce these tests (SingleDropEvap case with Nomura 471 setup):
PeleLMeX branch: https://github.com/baperry2/PeleLMeX/tree/test-spray-dep-species-fix
PelePhysics branch with these modifications and the heptaneABC mechanism: https://github.com/baperry2/PelePhysics/tree/dep-species-fix-test
PelePhysics branch without these modifications but with the heptaneABC mechanism: https://github.com/baperry2/PelePhysics/tree/spray-no-dep-fix-test

d-montgomery:
Thanks for providing the test cases. I understand the conundrum in the implementation now.  I've thought about it and the only  idea I have for improving efficiency is to order the `fuel_species` according to their `dep_fuel_species`.  This way the inner loop could be broken up into smaller chunks based on the `fdspec`. I think for now your approach is fine, but when we start using full GCxGC for the liquid species we may need to think of some clever work arounds.

d-montgomery (approved):

PR #593, opened 2025-07-11
Spray simplifications: Use PelePhysics unit conversions and EOS calls

baperry2:
The `define` function of the GasPhaseVals struct manually computes mw_mix and pressure. We can use EOS calls for that instead. 

To do that, some unit conversions are necessary, which motivates improvements to the unit conversions stuff in Spray. Instead of redefining unit conversions we can use the m2c stuff in PelePhysics utilities. We can also make all of this `static` and `constexpr` so it can be used without instantiating the SprayUnits struct.

In order to do that, we need to make everything possible constexpr in the m2c and c2m stuff.

### DIFF
--- a/Source/Spray/SprayFuelData.H
+++ b/Source/Spray/SprayFuelData.H
@@ -101,18 +101,6 @@ struct GasPhaseVals
       fluid_Y_dot[n] = 0.;
     }
   }
-
-  AMREX_GPU_DEVICE AMREX_FORCE_INLINE void define()
-  {
-    SprayUnits SPU;
-    mw_mix = 0.;
-    for (int n = 0; n < NUM_SPECIES; ++n) {
-      mw_mix += Y_fluid[n] / mw[n];
-    }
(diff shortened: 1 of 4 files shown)


d-montgomery (commented):
I tried running one of the SingleDropEvap tests in LMeX using this PR and am getting compilation errors:
~~~
In file included from ../../../Source/PeleLMeX.cpp:5:
In file included from PeleLMeX/Submodules/PelePhysics/Source/Spray/SprayParticles.H:4:
PeleLMeX/Submodules/PelePhysics/Source/Spray/SprayFuelData.H:42:39: 
error: no type named 'mks2cgs' in namespace 'pele::physics::utilities'
   42 | using m2c = pele::physics::utilities::mks2cgs;
~~~

baperry2:
My bad, I had fixed that but didn't pus the change. It should work now.

d-montgomery (approved):
This resolved the issue. Thanks!

PR #594, opened 2025-07-14
Utilities: Locate and square/circle intersection area

baperry2:
These functions are handy to have available for prob source files without needing to reimplement each time:

- `locate`: find the largest index that is less than an input value in a sorted array (used for interpolation). We have this implemented a few times across PeleC, PeleLMeX, and PelePhysics. I will collapse this to a single implementation for simplicity.
- `rectangle_circle_intersection_area`: useful for problems with round uniform-velocity inlets, by computing the intersection area with a cell we can scale the inlet velocity accordingly rather than a binary based on whether the cell center is within the inlet. This allows a mss flow to be specified and kept constant regardless of the resolution with which the inlet is resolved (even if it is smaller than a single cell!)

### DIFF
--- a/Testing/Exec/UtilityUnitTest/main.cpp
+++ b/Testing/Exec/UtilityUnitTest/main.cpp
@@ -1,0 +1,61 @@
+#include <AMReX.H>
+#include <AMReX_Print.H>
+#include "Utilities.H"
+
+int
+main(int argc, char* argv[])
+{
+  amrex::Initialize(argc, argv);
+
+  // Test Unit Conversions
+  namespace m2c = pele::physics::utilities::mks2cgs;
+  namespace c2m = pele::physics::utilities::cgs2mks;
+  AMREX_ALWAYS_ASSERT(m2c::Length(1.0) == 100.0);
+  AMREX_ALWAYS_ASSERT(c2m::Length(1.0, 2) == 1.0e-4);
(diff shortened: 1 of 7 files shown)


Testing/Exec/UtilityUnitTest/main.cpp
```diff
@@ -0,0 +1,61 @@
...
+#include "Utilities.H"
+
+int
+main(int argc, char* argv[])
+{
+  amrex::Initialize(argc, argv);
+
+  // Test Unit Conversions
+  namespace m2c = pele::physics::utilities::mks2cgs;
+  namespace c2m = pele::physics::utilities::cgs2mks;
```
d-montgomery: I think these aliases need to be moved above `main()`

baperry2:
For the utility unit tests that I added here, I had tried to set up a simpler GNUmakefile that would only build the relevant parts of the code, but for some reason I couldn't figure out, this was causing issues with CUDA. So I've reverted it to just use the standard `GNUmakefile` and `Make.PelePhysics`.

@marchdf and @jrood-nrel  - thoughts on the approach here for unit testing utilities? It's definitely not the prettiest but maybe it's good for now?

marchdf (approved):

marchdf:
I am fine with the slightly overly complicated gnumakefile for the unit tests. There's probably some needed amrex options that need to be set that aren't allowing you to remove as much as you would like...

d-montgomery (approved):

PR #595, opened 2025-07-15
Optimisation: exploiting symmetry of binary diffusion coefficients

ThomasHowarth:
This pull request changes how diffusion coefficients are evaluated to exploit the symmetric nature of the D_ij matrix. The NUM_SPECIES^2 loops and arrays are now over a size NUM_SPECIES*(NUM_SPECIES+1)/2, which roughly halves the computational cost of evaluating diffusion coefficients. 

Here is a tiny profile from the 2D flamesheet case in LMeX with the original implementation:
image

With the new implementation:
image

There's also a small tweak in the NASA polynomial computations that reuses the T2 and T3 values in computing the polynomial.


baperry2 (changes requested):
Looks good, this is a pretty solid performance gain. 

But before merging we also need the qss mechanisms to be regenerated: https://amrex-combustion.github.io/PelePhysics/Ceptr.html#batched-generation-of-pele-compatible-mechanisms

baperry2 (approved):
Now I get 0 diffs for `TranEval` for both `LiDryer` and `CH4_lean_qss`.

But it seems some extraneous changes came when regenerating the qss mechs due to the sympy stuff. I'll don't get those diffs when I regenerate so I'll push my versions. I think part but not all of the variability comes from us not being strict enough about the sympy version in pyproject.toml.

PR #599, opened 2025-07-22
Turbulent Forcings - Bug fix for high aspect ratio domains.

efhunt:
The implementation for breaking symmetry in high aspect ratio domains was not implemented in the addTurbVelForces despite being setup correctly. This PR just adds a small change to get it working.


efhunt:
I have closed so I can combine the bug fix with a second issues I discovered in my testing.

PR #601, opened 2025-07-29
add table lookup to utility unit test

baperry2:
Add simple test of table lookup capability that runs on host only


d-montgomery (approved):
Thanks for adding this example!

PR #602, opened 2025-07-29
Modularization of liquid fuel properties in spray

d-montgomery:
This PR is the beginning framework for generalizing the Spray module to work with different liquid property models (i.e. the existing model from PeleMP or a group contribution method). I have tested the changes using the [SingleDropEvap](https://github.com/AMReX-Combustion/PeleLMeX/tree/development/Exec/RegTests/SingleDropEvap) case in PeleLMeX and created a [draft PR](https://github.com/AMReX-Combustion/PeleLMeX/pull/544) for further testing. This is still a work in progress but I wanted to push the initial framework through now to avoid one giant PR later. 

Major changes:
1. Liquid properties and associated functions are stored in a struct that is selected based on a compile flag `SPRAY_GCM`. This struct (and soon others) are located in `SprayProperties.H`
2. Moved the `getInpCoef` and `getInpVal` functions to `SprayProperties.H` as these will change based on the selected method.

Minor changes
1. Mixture density (`rho_part`) is now a function. There are two versions of this function in order to be consistent with the existing PeleMP model. 
2. Renamed/removed some variables and added others for readability. Most notable is a new variable called `diff_liqLatent_gasEnth`, which replaces the old version of `latent` which wasn't really latent heat and was quite confusing at first.

Eventually all properties will be called as a function of temperature and users will have the option of using the GCM model from [FuelLib](https://github.com/nrel/fuellib). The necessary parameters will be read in at run time via an input file exported from FuelLib for Pele according to the updates in [this branch of FuelLib](https://github.com/dmontgomeryNREL/FuelLib-NREL/tree/export-crit-props).


d-montgomery:
The SprayTest in LMeX is catching some small differences that I wasn't seeing in the SingleDropEvap case. The issue begins after I merged the development branch into this branch in [#c5a41ea](https://github.com/AMReX-Combustion/PelePhysics/pull/602/commits/c5a41eaacd02adc6b921c9eff4403a569e41f4f2).  The SprayTest passes for commit [#22d8cd3](https://github.com/AMReX-Combustion/PelePhysics/pull/602/commits/22d8cd3bc3bf95ab0874a224e1ab148caec7a248) of this branch which was up to date with PelePhysics #598.

I checked out some of the recent commits from PelePhysics and this appears to be related to PR #595 (commit 037c048).

Source/Spray/BreakupSplash/AhamedSplash.H
```diff
@@ -225,21 +225,22 @@ droplet_splashing(
...
   // TODO: Determine correct method for handling multi-component liquids
   amrex::Real Tboil = 0.;
   for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
-    const int ysp = SprayComps::pstateY + spf;
+    Y_part[spf] = p.rdata(SprayComps::pstateY + spf);
     amrex::Real minT = amrex::min(T_part, cBoilT[spf]);
-    mu_part += p.rdata(ysp) * fdat.muL(minT, spf);
-    rho_part += p.rdata(ysp) / fdat.rhoL(minT, spf);
-    Tboil += p.rdata(ysp) * cBoilT[spf];
+    mu_part += Y_part[spf] * fdat.liqprops.mu_Li(minT, spf);
```
baperry2: how about making mixture functions for mu and Tboil (similar to rho)?

Source/Spray/BreakupSplash/WallFilm.H
```diff
@@ -66,10 +66,10 @@ calculateFilmSource(
   for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
     amrex::Real minT = amrex::min(T_film, cBoilT[spf]);
     Y_film[spf] = p.rdata(SprayComps::pstateY + spf);
-    rho_film += Y_film[spf] / fdat.rhoL(minT, spf);
-    cp_film += Y_film[spf] * fdat.cp[spf];
-    lambda_film += Y_film[spf] * fdat.lambdaL(minT, spf);
-    Tcrit += Y_film[spf] * fdat.critT[spf];
+    rho_film += Y_film[spf] / fdat.liqprops.rho_Li(minT, spf);
+    cp_film += Y_film[spf] * fdat.liqprops.cp[spf];
```
baperry2: Convert cp and critT to use accessor functions rather than directly accessing the data to make alternate (non-constant) implementations easier in the future?

Source/Spray/BreakupSplash/AhamedSplash.H
d-montgomery: That is the plan, I just didn't want to do too many things in this PR. Happy to add those if you want those included now. There is still a lot of generalization that needs to be done.

Source/Spray/SprayProperties.H
```diff
@@ -0,0 +1,208 @@
...
+
+// Original spray properties ported from PeleMP to PelePhysics
+struct MPLiqProps
+{
+  Real ref_T = 300.; // Reference temperature for liquid properties
+  GpuArray<Real, SPRAY_FUEL_NUM> critT;
+  GpuArray<Real, SPRAY_FUEL_NUM> boilT;
+  GpuArray<Real, SPRAY_FUEL_NUM> cp;
+  GpuArray<Real, SPRAY_FUEL_NUM> latent;
+  GpuArray<Real, SPRAY_FUEL_NUM> diff_liqLatent_gasEnth;
```
baperry2: This variable is confusing - we should figure out a more intuitive way to set this up at some point, but that can be a TODO

Source/Spray/SprayProperties.H
```diff
@@ -0,0 +1,208 @@
...
+    Real a = psat_coef[4 * spf];
+    Real b = psat_coef[4 * spf + 1];
+    Real c = psat_coef[4 * spf + 2];
+    Real d = psat_coef[4 * spf + 3];
+    return d * std::pow(10., a - b / (T + c));
+  }
+
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  Real latent_Li(const Real& T, const int spf) const
```
baperry2: I don't think the subscript L is needed here as the latent heat is a property of the phase transition rather than the liquid itself.

baperry2 (commented):
I think the overall structure is good and we can probably merge almost as is, then make some refinements/port more properties to this struct.

Before merging, we should think about the nomenclature for the functions for both individual component properties and mixture properties. I think we can make these a bit more clear. For individual components, I parse `_Li` as the surname rather than `liquid, component i`. For the mixture, I think `_part` is short for particle, but in principle this could be used to predict liquid properties other than for spray particles.

Perhaps we can just omit the `L` from e.g. `rho_Li`, as this is the `liqprops` struct so it should be clear that these are liquid properties. And then for mixture properties either just have no suffix/subscript, or use `_mix` instead of `_part`. In EOS for gas phase properties, we mostly use `i` to denote species properties and no subscript for mixture properties (e.g. TY2E and TY2Ei are mixture and component properties, respectively). But we're not quite consistent with that convention and perhaps it's a bit confusing anyway. I'm open to ideas on what would be the most intuitive.

Source/Spray/BreakupSplash/WallFilm.H
d-montgomery: Yes, this is exactly what I intend to do for `cp`.  Unless I'm mistaken, `critT` should always be constant.

Source/Spray/SprayProperties.H
d-montgomery: Are you referring to the `diff_liqLatent_gasEnth`? I agree that this is a confusing name. It was previously called `latent` even though it was `latent(T_ref) - h(T_ref)`, which was even more confusing.  

We could calculate `latent(T_ref) - h(T_ref)` each time the PeleMP model calculates the latent heat, but since these are constant I thought it made sense to only do it once.

Source/Spray/SprayProperties.H
d-montgomery: But you would keep `latent_i` in the `liqprops` struct right? This will change based on the selected model.

Source/Spray/BreakupSplash/WallFilm.H
baperry2: yeah critT is a constant property of the given species. I'd still prefer to set it up with accessor functions just for consistency (eventually all internal data to liqprops should be `private` and accessed through functions). With inlining, there's essentially no overhead to doing it that way.

Source/Spray/SprayProperties.H
baperry2: Yeah it should still be in liqprops because it comes into play when we have phase transition from liquid to gas.

Source/Spray/SprayProperties.H
baperry2: yes that's what I'm referring to. Maybe a balance between performance and readability would be to still precompute `h(T_ref)` but store it separately. Or change the combined variable name to `latent_minus_gasRefH_i` which is emphasizes the meaning better (physically what we care about is the latent heat and the gas reference enthalpy is just something we throw in for numerical convenience).

baperry2 (approved):

PR #603, opened 2025-07-30
Initialize Ddiag in Simple transport

baperry2:
ensure it is set at 0 before using `+=` operator


baperry2:
This appears to resolve some of the issues we are seeing in https://github.com/AMReX-Combustion/PeleLMeX/pull/545 but not all. Still some unexplained MLMG failures.

d-montgomery (approved):
Thanks for isolating the issue. I tested it with the latest development branch of LMeX and with #602 and the Spray test passes.

PR #604, opened 2025-07-31
Added eosparm object for spray manifold model

SreejithNREL:
Added an EosParm pointer to SprayData.
The pointer is initialised in the (newly added) overloaded function readSprayParams.
Added overloaded function spraySetup to populate species names using the eosparm object
Updated eos call in calculateSpraySource to read in manifold model
Updated eos call in updateParticles to read in manifold model
Added a dummy T2Cpi function for manifold model
Added a dummy EY2T function for manifold model


Source/Eos/Manifold.H
```diff
@@ -62,9 +62,7 @@ struct Manifold
   AMREX_FORCE_INLINE
   static void molecular_weight(amrex::Real* /*mw[]*/)
   {
-    amrex::Error(
-      "molecular_weight for species does not have significance for "
-      "Manifold EOS");
+    //DO NOTHING. MOLECULAR WEIGHT HAS NO SIGNIFICANCE IN MANIFOLD MODELS
```
baperry2: Leave as an error.

Source/Eos/Manifold.H
```diff
@@ -147,6 +145,14 @@ struct Manifold
     amrex::Error("RTY2Cv is not yet implemented for Manifold EOS");
   }
 
+  AMREX_GPU_HOST_DEVICE
+    AMREX_FORCE_INLINE
+    static void T2Cpi(const amrex::Real T, amrex::Real Cpi[NUM_SPECIES])
+    {
+	  // TODO: FIXME (T2Cpi need to be implemented for manifold model)
+      //amrex::Abort("\nT2Cpi is not yet implemented for Maniold model");
+    }
```
baperry2: Uncomment the `Abort`

Source/Eos/Manifold.H
```diff
@@ -172,6 +178,15 @@ struct Manifold
     // amrex::Error("TY2Cp is not yet implemented for Manifold EOS");
   }
 
+  AMREX_GPU_HOST_DEVICE
+    AMREX_FORCE_INLINE
+    static void
+    EY2T(const amrex::Real E, const amrex::Real Y[NUM_SPECIES], amrex::Real& T)
+    {
```
baperry2: Copy implementation of REY2T below rather than aborting (For this, only Y is actually used, so R and E are just ignored anyway).

Source/Spray/SpraySetup.cpp
```diff
@@ -70,6 +70,208 @@ getInpVal(
...
+      Print() << ", " << m_sprayFuelNames[i];
+    }
+#endif
+    Print() << std::endl;
+  }
+  Gpu::streamSynchronize();
+  ParallelDescriptor::Barrier();
+}
+
+
```
baperry2: Remove extra whitespace

Source/Spray/SpraySetup.cpp
```diff
@@ -70,6 +70,208 @@ getInpVal(
   }
 }
 
+
+
+void
+SprayParticleContainer::readSprayParams(int& particle_verbose,pele::physics::PeleParams<pele::physics::eos::EosParm<pele::physics::PhysicsType::eos_type>> *leosparm)
```
baperry2: Is it necessary to re-implement this whole function? I think you could probably just do the part where eosparm gets defined in the SpraySetup function, unless eosparm is needed before SpraySetup gets called.

if you do need to keep this function, this variable shouldn't be called leosparm, as that is typically used as the pointer to the actual eosparm data structure on device, not the PeleParams class. Instead of reimplementing the whole function, you could just do whatever you need to with eosparm and then call the version of the function without eosparm within this function.

Source/Spray/SpraySetup.cpp
```diff
@@ -105,7 +307,7 @@ SprayParticleContainer::readSprayParams(int& particle_verbose)
   // Must match the number specified at compile time
   const int nfuel = pp.countval("fuel_species");
   if (nfuel != SPRAY_FUEL_NUM) {
-    Abort("Number of fuel species in input file must match SPRAY_FUEL_NUM");
+    amrex::Print()<<"Warning! Number of fuel species in input file must match SPRAY_FUEL_NUM";
```
baperry2: I don't think we want to relax this?

Source/Spray/SprayParticles.H
```diff
@@ -242,6 +245,29 @@ public:
       pele::physics::TransportType> const* ltransparm,
     const amrex::Real spray_cfl_lev);
 
+  /// \brief Update particle data and solve gas-phase source terms (to be used with manifold model)
```
baperry2: I don't think this version gets defined or called anywhere - can you delete the declaration from the header?

Source/Spray/SprayParticles.cpp
```diff
@@ -327,10 +327,11 @@ SprayParticleContainer::updateParticles(
         refv.fillPtrs_d(rf_d);
       }
       auto* N_SB = N_SB_d.dataPtr();
+      auto const* leosparm = m_sprayData->eosparm;
       amrex::ParallelFor(Np, [=] AMREX_GPU_DEVICE(int pid) noexcept {
         ParticleType& p = pstruct[pid];
         if (p.id() > 0) {
-          auto eos = pele::physics::PhysicsType::eos();
+          auto eos = pele::physics::PhysicsType::eos(leosparm);
```
baperry2: use `fdat->eosparm`. This code will give you the host version, but this call is within the parallel for so you need the device version. Then you also won't need the line above defining leosparm.

Source/Spray/SpraySetup.cpp
```diff
@@ -314,6 +516,52 @@ SprayParticleContainer::spraySetup(const Real* body_force)
   ParallelDescriptor::Barrier();
 }
 
+void
+SprayParticleContainer::spraySetup(const Real* body_force, const pele::physics::eos::EosParm<pele::physics::PhysicsType::eos_type> *eosparm)
```
baperry2: I think this function is where you should define the pointers to eosparm in sprayData.

You should also be able to do this without having two separate versions of this function. Pass in two pointers to eosparm (one to the host version, one to the device version). Make them both optional arguments that default to nullptr. For non-manifold cases, everything should work fine passing `nullptr` around, so you wouldn't need to have an entire new implementation of the function.

Initially, set `m_sprayData->eosparm` to the device version, so that when `m_sprayData` gets copied to device `d_sprayData` ends up with the pointer to the device version. After the hostToDevice copy, change `m_sprayData->eosparm` to point to the host version of the eosparm. The goal should be that the host sprayData has the host eosparm and the device sprayData has the device eosparm.

Source/Transport/Manifold.H
```diff
@@ -38,8 +38,8 @@ struct ManifoldTransport
 
     if (wtr_get_lam) {
       // could be added to table if desired
-      amrex::Abort(
-        "Use of thermal conductivity is not supported for ManifoldTransport");
+      /*amrex::Abort(
```
baperry2: Leave as an abort. Or make it so that you can look up Lambda from the table if it exists (follow implementation for looking up mu - you'd have to add an idx_lambda in TransportParams for Manifold).

baperry2 (changes requested):
Please submit a parallel PR with the necessary changes in PeleLMeX.

Source/Eos/Manifold.H
```diff
@@ -172,6 +181,21 @@ struct Manifold
     // amrex::Error("TY2Cp is not yet implemented for Manifold EOS");
   }
 
+  AMREX_GPU_HOST_DEVICE
+  AMREX_FORCE_INLINE
+  void EY2T(
+    const amrex::Real /*E*/, const amrex::Real Y[NUM_SPECIES], amrex::Real& T)
+  {
+    // TODO: We need a corresponding function for this. For now, I am just
```
baperry2: Delete comment.

baperry2 (commented):
Please check and resolve the clang-tidy test errors.

Source/Spray/SprayParticles.cpp
```diff
@@ -327,10 +327,11 @@ SprayParticleContainer::updateParticles(
         refv.fillPtrs_d(rf_d);
       }
       auto* N_SB = N_SB_d.dataPtr();
+      auto const* leosparm = d_sprayData->eosparm;
```
baperry2: I don't think this line is needed

Source/Spray/SpraySetup.cpp
```diff
@@ -202,12 +205,16 @@ SprayParticleContainer::readSprayParams(int& particle_verbose)
...
+SprayParticleContainer::spraySetup(
+  const Real* body_force,
+  const pele::physics::eos::EosParm<pele::physics::PhysicsType::eos_type>*eosparm,
+	pele::physics::PeleParams<pele::physics::eos::EosParm<pele::physics::PhysicsType::eos_type>>* leosparm)
 {
 #if NUM_SPECIES > 1
   Vector<std::string> spec_names;
   pele::physics::eos::speciesNames<pele::physics::PhysicsType::eos_type>(
-    spec_names);
+    spec_names, (eosparm));
```
baperry2: remove unnecessary parentheses around eosparm

Source/Transport/Manifold.H
baperry2: This region should not be commented.

Submodules/amrex
baperry2, line 1: Why are there changes to amrex committed here?

baperry2:
Tested on CPU and GPU (cuda) from both PeleC and PeleLMeX. Every thing seems to be working

baperry2 (approved):

PR #607, opened 2025-08-04
Avoid machine precision subtraction issue in simple transport

baperry2:
In simple transport coefficient, the transport coefficient for species i is roughly proportional to $\sum Y_{j \neq i}/\sum X_{j \neq i}$. This leads to zero divided by zero numerical issues for Yi = Xi= 1. These are sidestepped by [adding an epsilon 1e-15 quantity](https://github.com/AMReX-Combustion/PelePhysics/blob/53972983afede14b8a50d4ccafa26e7e2f84d026/Source/Transport/Simple.H#L322) to all mass fractions just for purposes of the diffusion coefficient calculation to ensure no mass fraction is ever exactly unity in these calculations. 

This leads to numerical precision issues when we compute $\sum Y_{j \neq i}$ as $1 - Y_i$. Summing a bunch of quantities of O(1e-15) retains full floating point precision for sum, and therefore full floating point precision for the resulting diffusion coefficient. However, taking the difference between two O(1) quantities results in an O(1e-15) quantity with only O(1e-1) precision, and therefore results in a diffusion coefficient that can have O(1e-1) error, as seen in the EB-InflowBC test here: https://github.com/AMReX-Combustion/PeleC/pull/971. 

The proposed fix is to revert to computing $\sum Y_{j \neq i}$ instead of computing $1 - Y_i$, which unwinds part of the performance benefit from #595.

### DIFF
--- a/Source/Transport/Simple.H
+++ b/Source/Transport/Simple.H
@@ -146,9 +146,7 @@ struct BinaryDiff
   {
     const amrex::Real scale = Constants::PATM / (Constants::RU * Tloc);
 
-    amrex::Real sum_of_Yloc = 0.0;
     for (int i = 0; i < NUM_SPECIES; ++i) {
-      sum_of_Yloc += Yloc[i];
       Ddiag[i] = 0.0;
     }
 
@@ -160,14 +158,18 @@ struct BinaryDiff
                                tparm->fitdbin[four_idx_ij + 1] * logT[0] +
                                tparm->fitdbin[four_idx_ij + 2] * logT[1] +
                                tparm->fitdbin[four_idx_ij + 3] * logT[2];
-
         dbintemp = std::exp(-dbintemp);
         Ddiag[i] += Xloc[j] * dbintemp;
         Ddiag[j] += Xloc[i] * dbintemp;
       }
     }
     for (int i = 0; i < NUM_SPECIES; ++i) {
-      const amrex::Real term1 = sum_of_Yloc - Yloc[i];
+        amrex::Real term1 = 0;
+        for (int j = 0; j < NUM_SPECIES; ++j) {
+            if (i != j) {
+                term1 += Yloc[j];
+            }
+        }
       Ddiag[i] = tparm->wt[i] * term1 / Ddiag[i] * scale;
     }
   }
--- a/Source/Transport/Simple.H
+++ b/Source/Transport/Simple.H
@@ -164,12 +164,12 @@ struct BinaryDiff
       }
     }
     for (int i = 0; i < NUM_SPECIES; ++i) {
-        amrex::Real term1 = 0;
-        for (int j = 0; j < NUM_SPECIES; ++j) {
-            if (i != j) {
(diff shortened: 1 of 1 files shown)


Source/Transport/Simple.H
```diff
@@ -160,14 +158,18 @@ struct BinaryDiff
                                tparm->fitdbin[four_idx_ij + 1] * logT[0] +
                                tparm->fitdbin[four_idx_ij + 2] * logT[1] +
                                tparm->fitdbin[four_idx_ij + 3] * logT[2];
-
         dbintemp = std::exp(-dbintemp);
         Ddiag[i] += Xloc[j] * dbintemp;
         Ddiag[j] += Xloc[i] * dbintemp;
       }
     }
     for (int i = 0; i < NUM_SPECIES; ++i) {
```
ThomasHowarth, line 166: I think we could merge this for loop into the previous one, to avoid doing an extra N^2 loop, but also remove the if statement, i.e. 
1. we add term1=0 at the start of the loop
2. in the first nested loop (j < i) we add term1 += yloc[j]
3. Add another loop j=i+1, j < NUM_SPECIES which just adds the remaining part of term1
4. Construct Ddiag in the i loop.

ThomasHowarth (changes requested):

ThomasHowarth:
I had issues with pure species even before we added this change. I've been running pure hydrogen jets, and the diffusion solver was failing. I gradually increased the inflow mass fraction, and it was fine up to 99.9% hydrogen, but it would fail as soon as it hit 100%. Running with a fixed Lewis number ran fine, so I thought it was related to the mixture averaged formulation for pure species.

ThomasHowarth (changes requested):
Also, the same changes need to be made for the equivalent SRK function

baperry2:
@ThomasHowarth Good call on also making the changes for SRK.

Unless I'm missing something your proposed changes to the loop structure would result in the same number of FLOPS and I don't think we can move the full Ddiag within the first i loop so I did not implement that. 

One potential issue with the trace stuff for pure component diffusion is that it permanently modifies the input Yloc after the transport function is called, which is unexpected behavior. I `const` protected Yloc in this function and created a new temporary array for the modified version. I don't think this should change much in LMeX, which creates its own temporary arrays to store Y (after computing it from rhoY).

In general the whole process here when applied to pure components makes me a bit uneasy because we're essentially dividing by zero with a numerical hack that makes it sort of work. Not sure what a better approach would be though.

baperry2:
I'm going to merge this so we can get PelePhysics updates in PeleC going again. @ThomasHowarth we can revisit your suggestion on rearranging the loops later if you'd like

ThomasHowarth:
Sure thing @baperry2 - I opened a pull request into your [branch](https://github.com/baperry2/PelePhysics/pull/3) a while ago with my suggestions if you want to take a look

baperry2:
Sorry, somehow I completely missed that PR. I'll take a look.

PR #610, opened 2025-08-13
Add downstream tests

d-montgomery:
To avoid future headaches I'm proposing we add some downstream tests for LMeX and PeleC. I think the following should catch any potential issues from a PelePhysics PR:

- [x] clang-tidy for LMeX (EB-OFF only)
- [x] clang-tidy for PeleC 
- [x] FlameSheet for LMeX
- [x] PMF for PeleC
- [x]  Edit `workflow_dispatch` to only trigger for `development` branch

Let me know if you have any thoughts/suggestions @jrood-nrel , @marchdf and @baperry2.

When we are ready to merge, we'll want to edit the `workflow_dispatch` to read:
~~~
on:
  workflow_dispatch:
  push:
    branches: [development]
  pull_request:
    branches: [development]
~~~

I had to set the push branches to all to demonstrate the tests work as expected in this PR.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -1,9 +1,6 @@
 name: Downstream CI (PeleLMeX clang-tidy)
 
 on:
-  workflow_dispatch:
-  push:
-    branches: [development]
   pull_request:
     branches: [development]
 
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -62,10 +62,8 @@ jobs:
           echo "CCACHE_MAXSIZE=100M" >> $GITHUB_ENV
           echo "CTCACHE_DIR=~/.cache/ctcache" >> $GITHUB_ENV
 
-      - name: Install ccache & clang-tidy cache wrapper
+      - name: Install Ccache
         run: |
-          sudo apt-get update
-          sudo apt-get install -y xz-utils curl clang clang++ cmake
           wget https://github.com/ccache/ccache/releases/download/v4.8/ccache-4.8-linux-x86_64.tar.xz
           sudo curl https://raw.githubusercontent.com/matus-chochlik/ctcache/7fd516e91c17779cbc6fc18bd119313d9532dd90/clang-tidy-cache -Lo /usr/bin/clang-tidy-cache
           tar xvf ccache-4.8-linux-x86_64.tar.xz
@@ -73,7 +71,7 @@ jobs:
           sudo chmod +x /usr/bin/clang-tidy-cache
           mkdir -p ~/.cache/ctcache
 
-      - name: Cache (ccache & ctcache)
+      - name: Set Up Ccache
         uses: actions/cache@v4
         with:
           path: ~/.cache
@@ -96,7 +94,7 @@ jobs:
             -DCMAKE_CXX_COMPILER_LAUNCHER:STRING=ccache \
             ${{ github.workspace }}/PeleLMeX
 
-      - name: Check (build to run clang-tidy)
+      - name: Check
         working-directory: ${{ env.BUILD_DIR }}
         run: |
           cmake --build . --parallel ${{ env.NPROCS }} 2>&1 | tee -a clang-tidy-full-report.txt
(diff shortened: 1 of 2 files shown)


d-montgomery:
Looks like the changes I pulled in from #605 are causing the clang-tidy tests to fail.

baperry2:
Show's why this PR is needed haha! Can you apply the fixes as part of this PR?

baperry2 (approved):
Thanks for this. Looks good. But let's omit the PeleLMeX EB-ON clang-tidy just to keep the total number of tests down; I don't think it should provide anything the others don't (The PeleC build is EB-ON).

d-montgomery:
I removed the PeleLMeX EB-ON clang-tidy test and changed the `workflow_dispatch` to trigger for pushes and pull requests for development only. This should be good to go now.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,323 @@
...
+        working-directory: PeleC
+        run: |
+          set -eu -o pipefail
+
+          sudo add-apt-repository ppa:ubuntu-toolchain-r/test
+          sudo apt-get update
+
+          sudo apt-get install -y --no-install-recommends \
+              build-essential    \
+              gcc-10 \
```
jrood-nrel: Since GCC should already be available I don't think we need this.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,323 @@
...
+          mpirun -n 2 ./PeleC3d.gnu.MPI.ex example.inp max_step=2 amr.plot_int=-1 amr.check_int=-1 amrex.abort_on_unused_inputs=0 amr.n_cell=8 8 128
+
+  PeleLMeX-clang-tidy:
+    name: PeleLMeX clang-tidy
+    needs: [PeleLMeX-FlameSheet, PeleC-PMF]
+    runs-on: ubuntu-24.04
+
+    env:
+      BUILD_DIR: ${{ github.workspace }}/build-clang-tidy-lmex
+      use_eb: "OFF"
```
jrood-nrel: Doesn't look like this line is necessary.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,323 @@
...
+          export return=$(tail -n 1 clang-tidy-warnings.txt | awk '{print $2}')
+          exit ${return}
+
+  PeleC-clang-tidy:
+    name: PeleC clang-tidy
+    needs: [PeleLMeX-FlameSheet, PeleC-PMF]
+    runs-on: ubuntu-24.04
+
+    env:
+      BUILD_DIR: ${{ github.workspace }}/build-clang-tidy-c
```
jrood-nrel: I think we should be consistent with `BUILD_DIR` and either use it in all the jobs, or don't use it in all the jobs. I would recommend not using it and saving two lines.

.github/workflows/downstream.yml
```diff
@@ -0,0 +1,323 @@
...
+          make TPL COMP=gnu USE_MPI=TRUE
+          make -j 2 DIM=2 COMP=gnu USE_MPI=TRUE
+
+      - name: Run
+        working-directory: PeleLMeX/Exec/RegTests/FlameSheet/
+        run: |
+          mpirun -n 2 ./PeleLMeX2d.gnu.MPI.ex input.2d-regt amr.max_step=2 amr.plot_int=-1 amr.check_int=-1 amrex.abort_on_unused_inputs=1 amr.n_cell=32 64 
+
+  PeleC-PMF:
+    name: PeleC GNU@9.3 MPI Run [PMF]
```
jrood-nrel: I think it would be better to avoid listing GCC versions here.

jrood-nrel:
This is a good idea. Thanks for putting the work into this. I wish PeleC didn't require so much building.

baperry2:
Hmm for the Clang-Tidy here we don't really need to build everything. We just want clang-tidy to hit all the relevant pieces of PelePhysics. A combination like `MassCons`, `Soot-Flame`, and `Spray-EB` would probably be sufficient. Is there an easy way we could set it up to tell CMake to build only a particular subset of cases during the clang-tidy test?

d-montgomery:
Thanks for the suggestions @jrood-nrel. They should all be incorporated now.

> Hmm for the Clang-Tidy here we don't really need to build everything. We just want clang-tidy to hit all the relevant pieces of PelePhysics. A combination like `MassCons`, `Soot-Flame`, and `Spray-EB` would probably be sufficient. Is there an easy way we could set it up to tell CMake to build only a particular subset of cases during the clang-tidy test?

This is maybe a naive solution, but we could add a flag in CMake for LMeX and C, something like `PELE_UPSTREAM`, similar to what we do for the regression tests [here](https://github.com/AMReX-Combustion/PeleLMeX/blob/f82f7fa75b496ea0d556dab5e466fa6133b05f1c/Tests/CMakeLists.txt#L229) that includes the exact subset of tests we want to run from PelePhysics.

jrood-nrel:
I think we already do this https://github.com/AMReX-Combustion/PeleC/blob/2a001f023ef07fffbf9590e9ca6240170b34713d/CMakeLists.txt#L29

d-montgomery:
Thanks for pointing this out @jrood-nrel. I already have `DPELE_EXCLUDE_BUILD_IN_CI:BOOL=ON` in the PeleC clang-tidy test.  We could do something similar in LMeX, but those tests already take a lot less time to build/run than the PeleC test.

baperry2:
`PELE_EXCLUDE_BUILD_IN_CI` kind of does we what we want, but I think we could go much further here and exclude almost everything. @dmontgomeryNREL up to you if you want to add the `PELE_UPSTREAM` flag to PeleC/PeleLMeX or just be satisfied with what gets removed by `PELE_EXCLUDE_BUILD_IN_CI`

d-montgomery:
For now I think we should merge as is so we have some downstream testing setup. We can always refine in the future.

jrood-nrel (approved):

PR #612, opened 2025-08-25
Unit conversion fix

ThomasHowarth:
While working on [PeleLMeX#556](https://github.com/AMReX-Combustion/PeleLMeX/pull/556), I noticed that the RhoH conversion is the same for c2m and m2c instead of being inverted (and is what causes my current version to crash). This fixes the issue, and I hope it hasn't caused issues elsewhere.


d-montgomery (approved):
Dah, thanks for catching this @ThomasHowarth and I'm sorry for the headache it caused you. I don't see m2c::RhoH or c2m::RhoH called anywhere in the LMeX source code so hopefully this was an isolated issue.

PR #613, opened 2025-08-25
Generalize droplet mixture calculations

d-montgomery:
This PR creates mixture property functions within the MPLiqProps struct for the following:

- boiling point `boilT_mix`
- critical temp `critT_mix`
- specific heat `cp_mix`
- thermal conductivity `lambda_mix`
- dynamic viscosity `mu_mix`
- surface tension `sigma_mix`

Note that `rho_mix` was included in #602.  Managing the mixture averaged molecular weight of the spray is still a work in progress and not included here.

### DIFF
--- a/Source/Spray/BreakupSplash/AhamedSplash.H
+++ b/Source/Spray/BreakupSplash/AhamedSplash.H
@@ -228,10 +228,10 @@ droplet_splashing(
     Y_part[spf] = p.rdata(SprayComps::pstateY + spf);
   }
   // TODO: Determine correct method for handling multi-component liquids
-  amrex::Real Tboil = fdat.liqprops.boilT_mix(Y_part, cBoilT);
-  amrex::Real rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
-  amrex::Real mu_part = fdat.liqprops.mu_mix(Y_part, T_part, cBoilT);
-  amrex::Real sigma_part = fdat.liqprops.sigma_mix(Y_part, T_part, cBoilT);
+  const amrex::Real Tboil = fdat.liqprops.boilT_mix(Y_part, cBoilT);
+  const amrex::Real rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
+  const amrex::Real mu_part = fdat.liqprops.mu_mix(Y_part, T_part, cBoilT);
+  const amrex::Real sigma_part = fdat.liqprops.sigma_mix(Y_part, T_part, cBoilT);
   amrex::Real Tstar = fdat.wall_T / Tboil;
   const amrex::Real dia_part = p.rdata(SprayComps::pstateDia);
   const amrex::Real pmass = M_PI / 6. * rho_part * std::pow(dia_part, 3);
--- a/Source/Spray/BreakupSplash/AhamedSplash.H
+++ b/Source/Spray/BreakupSplash/AhamedSplash.H
@@ -231,12 +231,13 @@ droplet_splashing(
   amrex::Real Tboil = fdat.liqprops.boilT_mix(Y_part, T_part, cBoilT);
   amrex::Real rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
   amrex::Real mu_part = fdat.liqprops.mu_mix(Y_part, T_part, cBoilT);
+  amrex::Real sigma_part = fdat.liqprops.sigma_mix(Y_part, T_part, cBoilT);
   amrex::Real Tstar = fdat.wall_T / Tboil;
   const amrex::Real dia_part = p.rdata(SprayComps::pstateDia);
   const amrex::Real pmass = M_PI / 6. * rho_part * std::pow(dia_part, 3);
   // Weber number
   const amrex::Real We =
-    rho_part * dia_part * U0norm * U0norm / fdat.liqprops.sigma;
+    rho_part * dia_part * U0norm * U0norm / sigma_part;
   const amrex::Real Re_L = std::abs(U0norm) * dia_part * rho_part / mu_part;
   const amrex::Real Kv = std::sqrt(We * std::sqrt(Re_L));
   // Inclination of on-coming particle to surface in radians
--- a/Source/Spray/BreakupSplash/AhamedSplash.H
+++ b/Source/Spray/BreakupSplash/AhamedSplash.H
@@ -226,13 +226,14 @@ droplet_splashing(
   amrex::Real mu_part = 0.;
   amrex::Real rho_part = 0.;
   amrex::Real Y_part[SPRAY_FUEL_NUM] = {0.0};
-  // TODO: Determine correct method for handling multi-component liquids
-  amrex::Real Tboil = 0.;
   for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
     Y_part[spf] = p.rdata(SprayComps::pstateY + spf);
+  }
+  // TODO: Determine correct method for handling multi-component liquids
+  amrex::Real Tboil = fdat.liqprops.boilT_mix(Y_part, T_part, cBoilT);
+  for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
     amrex::Real minT = amrex::min(T_part, cBoilT[spf]);
     mu_part += Y_part[spf] * fdat.liqprops.mu_i(minT, spf);
-    Tboil += Y_part[spf] * cBoilT[spf];
   }
   rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
   amrex::Real Tstar = fdat.wall_T / Tboil;
--- a/Source/Spray/BreakupSplash/AhamedSplash.H
+++ b/Source/Spray/BreakupSplash/AhamedSplash.H
@@ -228,7 +228,7 @@ droplet_splashing(
     Y_part[spf] = p.rdata(SprayComps::pstateY + spf);
   }
   // TODO: Determine correct method for handling multi-component liquids
-  amrex::Real Tboil = fdat.liqprops.boilT_mix(Y_part, T_part, cBoilT);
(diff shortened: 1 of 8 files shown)


Source/Spray/BreakupSplash/AhamedSplash.H
```diff
@@ -223,24 +223,20 @@ droplet_splashing(
...
-  // TODO: Determine correct method for handling multi-component liquids
-  amrex::Real Tboil = 0.;
   for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
     Y_part[spf] = p.rdata(SprayComps::pstateY + spf);
-    amrex::Real minT = amrex::min(T_part, cBoilT[spf]);
-    mu_part += Y_part[spf] * fdat.liqprops.mu_i(minT, spf);
-    Tboil += Y_part[spf] * cBoilT[spf];
   }
-  rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
+  // TODO: Determine correct method for handling multi-component liquids
```
baperry2, line 230: Scary to see an unexplained message like this - but that's not something this PR needs to address (best to carry it forward if not resolved)

baperry2:
Can you explain what's going on with the different versions of each function with and without `cBoilT`? Could they be combined into a single implementation where the `cBoilT` argument defaults to a huge value to reduce repeated code?

Also, maybe we should be adding some documentation for the liqProps stuff now, before a documentation debt builds up too much.

Source/Spray/BreakupSplash/AhamedSplash.H
d-montgomery: Agreed... That was in there from the original PeleMP implementation so I left it for now. It's worth noting that the same mixture rule is used throughout, except for `rho_mix`.

d-montgomery:
> Can you explain what's going on with the different versions of each function with and without `cBoilT`? Could they be combined into a single implementation where the `cBoilT` argument defaults to a huge value to reduce repeated code?

The idea behind the overloaded functions was a result of some mixture density calls that used a single droplet temp [like this one](https://github.com/AMReX-Combustion/PelePhysics/blob/5be615f7b68d22f1019d96dd06c61c2708d1eb84/Source/Spray/SprayDerive.cpp#L76), while [many others](https://github.com/AMReX-Combustion/PelePhysics/blob/5be615f7b68d22f1019d96dd06c61c2708d1eb84/Source/Spray/Drag.H#L210) used a limited component-specific temperature `minT = amrex::min(T_part, cBoilT[spf])`. The later case was the most common and is used for all other mixture properties. 

> Also, maybe we should be adding some documentation for the liqProps stuff now, before a documentation debt builds up too much.

Everything we've done so far has been organizational and hasn't changed how the spray module is called from LMeX or C. I can add a brief overview of the liquid property functions used for the PeleMP method if you'd like.

d-montgomery:
I updated all the mixture functions to include a `null_ptr` for the third argument `cBoilT`, which defaults to `AMREX_REAL_MAX` if no pointer is passed in. This allows for a single function to handle both calls with and without `cBoilT`.  I tested the changes with the `SingleDropEvap` test in LMeX and everything works as expected. This should be good to go now.

Source/Spray/BreakupSplash/TABBreakup.H
```diff
@@ -38,29 +38,27 @@ updateBreakupTAB(
...
-  amrex::Real rho_part = 0.;
-  amrex::Real mu_part = 0.;
   amrex::Real Y_part[SPRAY_FUEL_NUM] = {0.0};
   for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
-    amrex::Real minT = amrex::min(T_part, cBoilT[spf]);
     Y_part[spf] = p.rdata(SprayComps::pstateY + spf);
-    mu_part += Y_part[spf] * fdat.liqprops.mu_i(minT, spf);
   }
-  rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
+  amrex::Real rho_part = fdat.liqprops.rho_mix(Y_part, T_part, cBoilT);
```
baperry2: Make all these `const` - you could probably go a bit crazy `const`ing things in these files. Start with anything applicable that you're adding/modifying here, and go as far beyond that as you want.

Source/Spray/SprayProperties.H
```diff
@@ -81,41 +81,108 @@ struct MPLiqProps
...
+    const amrex::Real* cBoilT = nullptr) const
+  {
+    amrex::Real mu = 0.;
+    for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
+      amrex::Real boil_temp =
+        (cBoilT != nullptr) ? cBoilT[spf] : AMREX_REAL_MAX;
+      mu += Y[spf] * mu_i(amrex::min(T, boil_temp), spf);
+    }
+    return mu;
+  }
```
baperry2, line 173: Here's an alternative implementation that is a bit more general and avoids the runtime conditional (used this as an excuse to try github copilot to see how it can help):

```
  template <typename RealArrayLike1, typename RealArrayLike2 = amrex::GpuArray<amrex::Real, SPRAY_FUEL_NUM>>
  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real mu_mix(
    const RealArrayLike1& Y,
    const amrex::Real T,
    const RealArrayLike2& cBoilT = []{
      amrex::GpuArray<amrex::Real, SPRAY_FUEL_NUM> arr; arr.fill(AMREX_REAL_MAX);
      return arr;
    }()) const
  {
    amrex::Real mu = 0.;
    for (int spf = 0; spf < SPRAY_FUEL_NUM; ++spf) {
      mu += Y[spf] * mu_i(amrex::min(T, cBoilT[spf]), spf);
    }
    return mu;
  }
```

baperry2:
The RealArrayLike templates are designed to work for any data type that returns amrex::Real from the [] operator. Should we be more explicit about this requirement in the code?

One strategy would to be to create a helper struct:
```
  // Helper trait to check if T has operator[] returning amrex::Real& or const amrex::Real&
  template <typename T>
  struct is_real_array_like {
    template <typename U>
    static auto test(U* u) -> std::integral_constant<
      bool,
      std::is_same<decltype((*u)[0]), amrex::Real&>::value ||
      std::is_same<decltype((*u)[0]), const amrex::Real&>::value>;
    template <typename>
    static auto test(...) -> std::false_type;
    static constexpr bool value = decltype(test<T>(0))::value;
  };
```
And then in each template include a static assert:
```
static_assert(is_real_array_like<RealArrayLike1>::value,
               "RealArrayLike1 must have an operator[] that returns const amrex::Real&");
```

This has no effect on code behavior or what can compile, but makes expectations more explicit/might increase readability of the code. But it also adds significant verbosity and can potentially add confusion rather than reduce it if the reader isn't as familiar with C++.

Source/Spray/SprayProperties.H
d-montgomery: I like this idea and have implemented something similar now, but I'm starting to wonder if we're over thinking this a bit. The only function this is really needed for is the mixture density, as there are calls requiring `rho_mix(Y, T)` and `rho_mix(Y, T, cBoilT)`.  All other mixture calls require `Y, T, cBoilT`.

I think it makes sense to implement something flexible for `rho_mix`, but for the other functions a simpler implementation is more readable. 

Thoughts?

baperry2: Up to you. I don't think we get that much simplification (just removing the defaults from the second template argument and cboilT function argument), but if the version without cBoilT isn't actually meaningful there's no point in supporting it.

d-montgomery: Ya, I can't see a case with the current PeleMP spray module where the other mixture functions are called without `cBoilT` so I'm going to simplify those calls. `cBoilT` won't be required with the GCM implementation, but the function calls will still need all three parameters to satisfy both models.

Source/Spray/SprayProperties.H
```diff
@@ -81,41 +81,116 @@ struct MPLiqProps
...
+      Tboil += Y[spf] * cBoilT[spf];
+    }
+    return Tboil;
   }
 
   template <typename RealArrayLike>
-  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real rho_mix(
+  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real cp_mix(
     const RealArrayLike& Y,
     const amrex::Real& T,
```
baperry2: Pass T by value instead of by reference,. here and below

Source/Spray/SprayProperties.H
```diff
@@ -81,41 +81,116 @@ struct MPLiqProps
...
-  // amrex::Real&
+  // Mixture functions use RealArrayLike to allow for different types of arrays
+  // with an [] operator that returns amrex::Real (or convertible to
+  // amrex::Real). cBoilTLike is used in templated functions to allow both
+  // amrex::GpuArray and amrex::Real* types for boiling temperature arrays.
+
+  // Helper function to get effective temperature considering boiling point
+  template <typename cBoilTLike>
+  AMREX_GPU_HOST_DEVICE AMREX_FORCE_INLINE amrex::Real effectiveT(
+    const amrex::Real& T, const int spf, const cBoilTLike& cBoilT) const
```
baperry2: I don't like the template typename "cBoilTLike" - the reader probably has no idea what that means and it's redundant with the variable name. `cBoilT` just has to be `RealArrayLike` as defined here - it could be a GpuArray of amrex::Real, a pointer to amrex::Real, or in principal even something else like amrex::CellData.

I'm also not wild about this helper function in general because it doesn't actually simplify the code, but I'm fine with leaving it in if you think it makes the intent of the code more clear.

d-montgomery: That's fair. I removed the helper function and replaced `cBoilTLike` with `RealArrayLike2` per your previous implementation idea.

baperry2 (approved):

PR #614, opened 2025-08-28
Spray changes for Multicomponent and Manifold Stuff

baperry2:
The goal here is to update the documentation to reflect what needs to be done for these, then implement those changes in the code.

For now this just looks at modifications for improving the capability to have many liquid phase components dump into a single gas phase component. Perhaps some refinement on the approach is still needed, but what's in there now should have the property that if all of the many components have identical properties, the result is the same as a single component of those properties (the current code doesn't exactly satisfy this despite improvements in #592)


baperry2:
Superseded by #633, which implements the changes described in this documentation update for many-to-one liquid-to-gas mapping of multicomponent sprays.

The manifold changes were already mostly merged in #620, with potential additional changes to come in #628.

PR #628, opened 2025-10-15
Spray with manifold model implementation

SreejithNREL:
First working (compiles, runs) spray evaporation model coupled with manifold combustion model. 
Validation is still in progress for the single droplet evaporation case.


Source/Spray/Drag.H
```diff
@@ -250,9 +250,8 @@ struct calcVaporState<pele::physics::eos::Manifold>
     eos.Y2WBAR(Y_skin, mw_skin);
     mw_skin *= SprayUnits::mass_conv;
 
-    // T_skin is just a dummy here
+    // Cp is already in kJ/kg-K so do not use any conversion unit
     eos.TY2Cp(T_skin, Y_skin, cp_skin);
```
baperry2, line 254: I believe this unit conversion should be necessary. Any quantity coming from a PelePhysics EOS call should be in CGS units. If it isn't, that's a problem with the table generation rather than this call.

baperry2:
Closing, will open a new PR with relevant changes on a clean branch.

PR #632, opened 2025-10-23
Cleanup some spray stuff

marchdf:
No need to pass pointers around, or keep them as member vars. This is going to require a companion PR to the other codes.


jrood-nrel (approved):

marchdf:
The CI passes in the pelec tests so I think we can merge.

PR #634, opened 2025-10-27
Update magma for ellis

marchdf:


marchdf:
not needed I think. If the module is loaded, it will find openblas.

PR #635, opened 2025-10-28
Fix uninitialized drag warnings

marchdf:


jrood-nrel:
Should we initialize all the ones that aren't reporting warnings as well?

marchdf:
good idea, done.

jrood-nrel (approved):

baperry2:
Not sure what the best strategy is on these. 

Initialization like
```
amrex::GpuArray<amrex::Real, SPRAY_FUEL_NUM> L_fuel{{0.0}};
```
initializes the first element to the bracketed value and all other elements to 0 regardless of the bracketed value. Which I guess is fine but IMO is actually more bug prone and confusing than leaving them uninitialized where not required (as someone may be tempted to follow this pattern using a value other than zero, leading to unanticipated behavior, don't ask me how I know...). 

I prefer the strategy of declaring unintialized and calling the `.fill()` routine to initialize with the desired value if needed, which I think is more readable/less likely to lead to bugs.

marchdf:
hmmm interesting. I can call a .fill on these

marchdf:
Ok I did the fill thingy

baperry2 (approved):

PR #639, opened 2025-10-29
Spray validation updates for Manifold EOS

baperry2:
Supersedes #628

Correct the skin properties using a scaling based on Sutherland's law for Transport properties to better account for actual skin temperature. Use Cp_fuel where needed (rather than previously using Cp_mixture). Ideally we should apply some temperature correction on Cp_fuel as well, but for now the validation is pretty good without it for the Wong-Lin case (which had poor validation before).

Requires https://github.com/NREL/cmlm/pull/23

Here's the Wong-Lin case with these updates (both actually using MP/antoine, the GCM label is using Manifold EOS):
Screenshot 2025-10-29 at 3 53 19 PM

TODO:
- [ ] Document spray+manifold capability approach and validation
- [x] Corresponding modification to https://github.com/AMReX-Combustion/PeleLMeX/pull/571
- [ ] Fix indexing for diffusion coefficients in `Drag.H`

### DIFF
--- a/Source/Eos/EosParams.H
+++ b/Source/Eos/EosParams.H
@@ -47,6 +47,7 @@ struct EosParm<Manifold>
   int idx_T{0};
   int idx_Wdot[MANIFOLD_DIM];
   int idx_Cp{-1};
+  int idx_Cp_fuel{-1};
   int is_variance_of[MANIFOLD_DIM];
   int verbose{1};
   bool has_mani_src{true};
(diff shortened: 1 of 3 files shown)


baperry2:
My inclination here is to defer the remaining TODOs to future PRs so this and the LMeX PR can be merged to make way for #633.

d-montgomery (approved):

d-montgomery (approved):
Great idea to use Sutherlands here!

PR #648, opened 2026-01-16
Use amrex min and max

jrood-nrel:


baperry2 (approved):
Step toward better single precision support.

PR #650, opened 2026-01-28
Spray data issue on hip

SreejithNREL:
A draft bug fix for resolving host-device declaration/access error on HIP.

### DIFF
--- a/Source/Spray/SprayParticles.H
+++ b/Source/Spray/SprayParticles.H
@@ -45,8 +45,8 @@ public:
 
   static void SprayCleanUp()
   {
-    //delete m_sprayData;
-    //amrex::The_Arena()->free(d_sprayData);
+    // delete m_sprayData;
+    // amrex::The_Arena()->free(d_sprayData);
     d_sprayData = nullptr;
     m_sprayData = nullptr;
   }
--- a/Source/Spray/SprayParticles.H
+++ b/Source/Spray/SprayParticles.H
@@ -45,9 +45,10 @@ public:
 
   static void SprayCleanUp()
   {
-    delete m_sprayData;
-    // amrex::The_Device_Arena()->free(d_sprayData);
+    //delete m_sprayData;
+    //amrex::The_Arena()->free(d_sprayData);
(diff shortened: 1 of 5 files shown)


jrood-nrel:
Why is the executable in CI `PeleLMeX2d.gnu.MPI.drm19.ex` all of a sudden?

baperry2:
> Why is the executable in CI `PeleLMeX2d.gnu.MPI.drm19.ex` all of a sudden?

Changed in #590 for the flamesheet case using the USERSuffix GNUmake option. It's helpful for keeping things straight when compiling with various different mechanisms.

Source/Spray/SprayParticles.H
```diff
@@ -46,7 +46,8 @@ public:
   static void SprayCleanUp()
   {
     delete m_sprayData;
-    amrex::The_Arena()->free(d_sprayData);
+    // amrex::The_Device_Arena()->free(d_sprayData);
```
jrood-nrel: Feels like we need some kind of free happening if we have the allocate still happening?

SreejithNREL: For some reason, the freeing of the spray data structure ```d_sprayData``` appears to occur after the AMREX finalize call, leading to the warning at the end of the PeleLMeX run.

```
pure virtual method called
terminate called without an active exception
```

I am thinking of freeing the memory at the destructor in PeleLMeX itself as:
```
PeleLM::~PeleLM()
{
..
#ifdef PELE_USE_SPRAY 
  SprayPC.reset();
  GhostPC.reset();
  VirtPC.reset();
#endif
..
}
```

These changes resolved the warning. I will raise the changes in PeleLMeX.cpp as a separate PR

baperry2:
@SreejithNREL: can you provide a bit more explanation for the changes you've made here? 

This looks weird to me, especially setting to nullptr instead of delete and relying on the external codes to free memory (likely to be more bug-prone) rather than doing it internally.

d-montgomery:
@SreejithNREL any updates on this PR?

PR #653, opened 2026-02-16
Compute element fractions in EOS

baperry2:
- Moves some functions associated with element atomic weights from `mechanism.cpp` to `mechanism.H` and makes them `AMREX_GPU_HOST_DEVICE` (already fully device compatible, they just need the tag)
- Adds a function only supported in FUEGO and SRK EOS that computes element fractions of a mixture using these newly device-capable functions
- regenerates all mechanisms with CEPTR to account for these changes


jrood-nrel (approved):

d-montgomery (approved):
Thanks for adding this @baperry2. We should make sure to add a new version with this PR since all `CustomMechanisms` will need to be regenerated with the updated ceptr script.

PR #656, opened 2026-02-18
Move a print function outside of a parallelfor

jrood-nrel:
Looking at C++20, it seems capturing `this` in lambdas isn't allowed, so we move this `print_final_stats` outside of the `ParallelFor`. I don't like these prints inside the ParallelFors regardless.


baperry2 (approved):
Agree. amrex::Print() shouldn't be within a ParallelFor

PR #661, opened 2026-03-10
Refactor PeleRad CMake tests

jrood-nrel:
In the nightly testing I have been seeing `CMake can not determine linker language for target` errors. This is because the sources for the `PeleRad` library were only headers and CMake was having trouble deciding if they were CXX or C. Rather than care about this, I decided to not list the headers as sources in this case so that we could just avoid using a `PeleRad` library and simplify the PeleRad tests at the same time.


baperry2:
The Rad tests currently run as part of the PeleLMeX CI, which doesn't really make sense. Can we move them over so they run here instead of there as part of this?

jrood-nrel:
That sounds nice, but that probably means we need PelePhysics to be its own CMake project, which isn't trivial.

baperry2:
Ok. Can you submit a parallel PeleLMeX PR so we can be sure the Rad tests still run properly before we merge this then?

jrood-nrel:
I think the PeleLMeX PR is passing the tests now.

baperry2 (approved):
The PeleLMeX tests that are failing to run in debug mode seem to be a separate issue, so this should be good to go now.

PR #667, opened 2026-04-01
Add option to loop turbinflow planes in time

d-montgomery:
This PR adds a flag `turbinflow.__.time_periodic`, which enables users to loop turbinflow planes periodically in time.


baperry2:
Thanks for doing this, I was starting on an overcomplicated approach!

baperry2 (approved):

PR #669, opened 2026-04-09
Fix volume used  in 2D spray_density calculation  

d-montgomery:
In 2D, the spray source-term coupling in `SprayParticles.cpp` treats the domain as a thin slab with `Lz = dx[0]`, using `V_slab = dx[0] * dx[0] * dx[1]`. The `computeDerivedVars` computation in `SprayDerive.cpp` was using only the 2D cell area `dx[0] * dx[1]`, causing `spray_density` and `spray_vol_frac` to be overestimated by a factor of `dx[0]` in plotfiles. This applies the same correction to cell volume in `computeDerivedVars` so diagnostic fields are consistent with the source-term coupling.

### DIFF
--- a/Source/Spray/SprayDerive.cpp
+++ b/Source/Spray/SprayDerive.cpp
@@ -18,7 +18,11 @@ SprayParticleContainer::computeDerivedVars(
   const RealVect dxi(AMREX_D_DECL(dxiarr[0], dxiarr[1], dxiarr[2]));
   const RealVect dx(AMREX_D_DECL(dxarr[0], dxarr[1], dxarr[2]));
   const RealVect plo(AMREX_D_DECL(ploarr[0], ploarr[1], ploarr[2]));
-  const Real cell_vol = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
+  Real cell_vol = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
+#if AMREX_SPACEDIM == 2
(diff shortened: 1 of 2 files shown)


Source/Spray/SprayDerive.cpp
```diff
@@ -18,7 +18,11 @@ SprayParticleContainer::computeDerivedVars(
   const RealVect dxi(AMREX_D_DECL(dxiarr[0], dxiarr[1], dxiarr[2]));
   const RealVect dx(AMREX_D_DECL(dxarr[0], dxarr[1], dxarr[2]));
   const RealVect plo(AMREX_D_DECL(ploarr[0], ploarr[1], ploarr[2]));
-  const Real cell_vol = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
+  Real cell_vol = AMREX_D_TERM(dx[0], *dx[1], *dx[2]);
```
baperry2: This can be simplified:
```suggestion
  // Spray model in 2D assumes narrow domain in z-direction (Lz = dx)
  Real cell_vol = AMREX_D_PICK(dx[0]*dx[0]*dx[0], dx[0]*dx[1]*dx[0], dx[0]*dx[1]*dx[2]);
```

d-montgomery: Good call. I'll add this to the `SprayParticles.cpp` too.

baperry2 (approved):
Looks good once formatting is fixed.

PR #671, opened 2026-04-23
Spray: only fully vaporize if dropT > max(boilT)

d-montgomery:
In `Drag.H` particles were considered fully vaporized when the droplet temperature exceeded the estimated mixture boiling temp. However, for multicomponent fuels this lead to premature total vaporization. The existing SingleDrop tests have relatively low temperatures so this never came up.


d-montgomery:
After some testing with the new Burger et al. validation case [https://github.com/AMReX-Combustion/PeleLMeX/pull/660](https://github.com/AMReX-Combustion/PeleLMeX/pull/660), I am okay with removing the droplet temperature check entirely and relying on mass only.

SreejithNREL (approved):
The code fix looks good to me. Using 'boilT_mix' as a hard cut-off criterion is fragile for multicomponent fuels, where preferential evaporation can result in  individual component temperature shooting above boilT_mix while the droplet is active. Nice catch!
I see the following code changes:
1. fixed the criterion to new_mass > min_mass alone
2. removed old, unused variable Tboil, and
3. updated documentation (including mentioning why the 50 bar case is not being evaluated at present)
All looks good to me.

Approving.

baperry2:
What happens now when the droplet temperature exceeds the boiling point in a single component case? This is a semi-nonphysical condition (superheating is possible, but shouldn't happen to excessive amounts), so just want to double check that the code won't crash/do something crazy after this modification.

d-montgomery:
> What happens now when the droplet temperature exceeds the boiling point in a single component case? This is a semi-nonphysical condition (superheating is possible, but shouldn't happen to excessive amounts), so just want to double check that the code won't crash/do something crazy after this modification.

I ran the WongLin and Nomura tests out until the drops vaporized without issue. In both cases the gas temperature is much higher than the boiling point of droplet. Here are the results from the WongLin test:
WongLin

PR #679, opened 2026-05-21
Make CEPTR code generation fully deterministic

terencelehmann:
## Summary

CEPTR code generation can produce trivially different (but equivalent) output
across runs when `PYTHONHASHSEED` varies, causing the CI diff-check to flag
spurious changes, as mentioned in #659 . This PR aims to fix the sources of non-determinism in the generator:

- `intersection()` in `utilities.py`: replace `list(set(...))` with `sorted(...)`
- `identify_qssa_coupling()` in `qssa_reduction.py`: same fix for the deduplication step
- `free_symbols` iterations in `species_info.py` and `symbolic_math.py`: sort by `str` before iterating
- CI `CEPTR` job: set `PYTHONHASHSEED: 0` to fix SymEngine CSE variable numbering,
  which otherwise varies with Python's hash randomization

All four existing tests pass.

---

Co-authored by: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,8 @@ jobs:
       matrix:
         python-version: ['3.10', '3.11']
         poetry-version: ['1.4.2']
+    env:
+      PYTHONHASHSEED: 0
     defaults:
       run:
         working-directory: ${{github.workspace}}/Support/ceptr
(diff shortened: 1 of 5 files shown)


drummerdoc:
Thanks for working on this, Terance.  If you get a chance, and can figure a way to do it, is it possible to add a CI test that confirms that the results will be deterministic?

PR #687, opened 2026-08-10
Less ci

d-montgomery:
This limits the amount of CI required when changes are only made in Docs/ files or Utility python scripts.


baperry2 (approved):
Good idea to not waste compute cycles on unneeded testing.

d-montgomery:
Thanks @baperry2. I think I need to merge this PR to test if it really worked. I'll add some similar `paths-ignore` in LMeX and C.

